PR #6, opened 2018-07-21
Adding Murax SoC (and misc improvements)

### LINKED ISSUE #4: Add Murax SoC as test example
From https://github.com/SpinalHDL/VexRiscv#murax-soc
> Murax is a very light SoC (it fits in an ICE40 FPGA) which can work without any external components.
>
> Depending the CPU configuration, on the ICE40-hx8k FPGA with icestorm for synthesis, the full SoC has the following area/performance :
> * RV32I interlocked stages => 51 Mhz, 2387 LC 0.45 DMIPS/Mhz
> * RV32I bypassed stages => 45 Mhz, 2718 LC 0.65 DMIPS/Mhz

This should be small enough to fit on the `up5k` and the `hx8k` (and the `lp8k`) making a good target which hits all our boards?

We should probably generate the verilog once and push it to a repo so people don't have to use the SpinalHDL stuff.

mithro:
* Adding Murax SoC (fixes #4).
 * Tested with vpr and nextpnr on hx8k and ct256.
 * Added support for data files (needed for the SoC memory init).
 * Cleanup some whitespace and argument stuff.
 * Small document improvements.


fpgaperf.py
```diff
@@ -381,10 +387,6 @@ def __init__(self):
         Toolchain.__init__(self)
         self.toolchain = 'vpr'
 
-    # FIXME: hack until vpr is fixed (it has hx8k-cm81 instead of lp8k-cm81)
-    def device_workaround(self):
```
JohnDMcMaster: Is this fixed in VPR now?

mithro: Yeap!

README.md
```diff
@@ -93,6 +94,18 @@ spnr
 vpr
 ```
 
+You can check if you have the toolchain environments correctly installed as
+follows:
+```
+$ python3 fpgaperf.py --check-env --toolchain vpr
```
mcmasterg: What are your thoughts on --check-env? I was toying around with this since the dependencies could get quite onerous for this project. Is this useful and/or the right thing to do?

fpgaperf.py
```diff
@@ -817,9 +823,11 @@ def list_seedable():
     for t in get_seedable():
         print(t)
 
-def check_env():
+def check_env(to_check):
```
mcmasterg: if to_check is invalid, does this silently fail?

fpgaperf.py
```diff
@@ -851,9 +859,9 @@ def main():
     parser.add_argument('--device', default='hx8k', help='Device within family')
     parser.add_argument('--package', default=None, help='Device package')
     parser.add_argument('--strategy', default='default', help='Optimization strategy')
-    parser.add_argument('--toolchain', help='Tools to use')
+    parser.add_argument('--toolchain', help='Tools to use', choices=get_toolchains())
     parser.add_argument('--list-toolchains', action='store_true', help='')
-    parser.add_argument('--project', help='Source code to run on')
+    parser.add_argument('--project', help='Source code to run on', choices=get_projects())
     parser.add_argument('--list-projects', action='store_true', help='')
     parser.add_argument('--seed', default=None, help='31 bit seed number to use, possibly directly mapped to PnR tool')
```
mcmasterg: ideally we should also choices=get_seedable()

README.md
mithro: I was *super* useful for me. I really liked it.

fpgaperf.py
mithro: Fixed.

fpgaperf.py
mithro: `--seed` takes an integer? Are you saying check that `--seed` is only used on projects which have `get_seedable()`?

README.md
mithro: It seems great to me!

fpgaperf.py
mithro: Fixed.

fpgaperf.py
mcmasterg: er yeah nvm

PR #26, opened 2019-10-09
Add initial Vivado support

kgugala:
This PR adds initial support for Vivado. It uses edalize for handling the tool. Edalize from  https://github.com/olofk/edalize/pull/54 should be used with this code.

Most of the resources usage are read from utilization report, currently the reported ones are limited to the set reported so far for different tools. This should be probably expanded.

Also, max_frequency is not yet extracted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ $ git submodule init
 $ git submodule update
 ```
 You'll need the following tools in your path:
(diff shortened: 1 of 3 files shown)


mithro:
Please update the documentation in the README too....

PR #36, opened 2019-12-03
Refactor toolchain handling and use edalize for icestorm

kgugala:
This commit moves toolchain, utils and icestorm targets (arachne and nextpnr) out of main script. This way the code is more readable.

Also icestorm targets are now using edalize.

Signed-off-by: Karol Gugala <kgugala@antmicro.com>


toolchain.py
```diff
@@ -0,0 +1,238 @@
...
+import datetime
+from utils import Timed
+
+class Toolchain:
+    # List of supported carry modes
+    # Default to first item
+    carries = None
+    strategies = None
+
+    '''A toolchain takes in verilog files and produces a .bitstream'''
```
mithro: This comment is in the wrong place.

utils.py
```diff
@@ -0,0 +1,16 @@
+import time
+
+class Timed():
```
mithro: Unneeded `()`

kgugala:
> Does this needs to be reformatted to match pep8 / other code style?

yes, it does. Do we want pep8 or maybe yapf?

mithro:
I logged https://github.com/SymbiFlow/fpga-tool-perf/issues/37 to track the formatting issue. Let's not block this on that.

toolchain.py
kgugala: fixed

utils.py
kgugala: fixed

PR #38, opened 2020-02-17
Update Vivado flow

kgugala:
This PR extracts Vivado flow from the main script. Also Yosys -> Vivado flow has been added.

This code requires edalize from this PR: https://github.com/olofk/edalize/pull/103

### DIFF
--- a/vivado.py
+++ b/vivado.py
@@ -1,0 +1,129 @@
+import os
+import subprocess
+import time
+import collections
+import json
+import re
+import shutil
+import sys
+import glob
+import datetime
+import asciitable
+import edalize
+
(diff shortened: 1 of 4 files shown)


vivado.py
```diff
@@ -0,0 +1,157 @@
...
+            "IOB" : str(iob),
+            }
+        return ret
+
+    def versions(self):
+        return self.backend.get_version()
+
+class VivadoYosys(Vivado):
+    '''Vivado PnR + Yosys synthesis'''
+    carries = (False, False)
```
mithro: Newline....

kgugala: fixed

PR #39, opened 2020-02-20
Add support SymbiFlow Xilinx flow

kgugala:
This PR adds support for Xilinx Symbiflow flow.
It requires edalize from: https://github.com/SymbiFlow/edalize/tree/symbiflow
The code here assumes SymbiFlow is installed using install routine from https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1252


symbiflow.py
```diff
@@ -0,0 +1,232 @@
...
+                print("Adding pcf file", self.pcf)
+                self.files.append({'name' : os.path.realpath(self.pcf), 'file_type': 'SDC'})
+
+            chip = self.family + self.device
+
+            self.edam = {
+                'files' : self.files,
+                'name' : self.project_name,
+                'toplevel' : self.top,
+                'tool_options' : {'symbiflow' :
```
mithro: Wrapping of this dictionary is a bit weird.

symbiflow.py
```diff
@@ -0,0 +1,232 @@
...
+            if l.find('Version:') == 0:
+                version = l
+            if l.find('Revision:') == 0:
+                revision = l
+        assert version is not None
+        assert revision is not None
+        return version + ', ' + revision
+
+    def versions(self):
+        return {
```
mithro: Wrapping here is a bit weird.

symbiflow.py
```diff
@@ -0,0 +1,232 @@
...
+            'vpr': VPR.vpr_version(),
+            }
+
+    @staticmethod
+    def seedable():
+        return True
+
+    @staticmethod
+    def check_env():
+        return {
```
mithro: Wrapping here is a bit weird.

kgugala: fixed

symbiflow.py
kgugala: fixed

symbiflow.py
kgugala: fixed

PR #41, opened 2020-02-26
Litex Linux test

kgugala:
This PR adds Linux/Litex test.

For symbiflow toolchain it requires binary toolchain installed from: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1252 and edalize from https://github.com/SymbiFlow/edalize/tree/symbiflow


mithro:
LGTM.

PR #57, opened 2020-04-10
Add nextpnr

acomodi:
This PR adds nextpnr to the flow, as well as a new blinky test that works on the RGB leds of the arty platform.

The blinky test is targetted, for now for:

- VPR;
- Nextpnr.

This PR depends on https://github.com/SymbiFlow/edalize/pull/28

Follow up PRs will include:

- QoR extraction for nextpnr
- Integration of some of the pre-existing tests to address nextpnr

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,3 +20,6 @@
 [submodule "third_party/prjxray"]
 	path = third_party/prjxray
 	url = https://github.com/SymbiFlow/prjxray.git
+[submodule "third_party/nextpnr"]
+	path = third_party/nextpnr
+	url = https://github.com/daveshah1/nextpnr-xilinx.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,6 +20,3 @@
 [submodule "third_party/prjxray"]
 	path = third_party/prjxray
 	url = https://github.com/SymbiFlow/prjxray.git
-[submodule "third_party/nextpnr"]
-	path = third_party/nextpnr
-	url = https://github.com/daveshah1/nextpnr-xilinx.git
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ SHELL = bash
 PWD = $(shell pwd)
 INSTALL_DIR = ${PWD}/third_party/install
 
+NEXTPNR_DEVICES = xc7a35tcsg324-1
+
 all: format
 
 conda:
--- a/conf/environment.yml
+++ b/conf/environment.yml
@@ -8,6 +8,7 @@ dependencies:
   - yosys
   - yosys-plugins
   - vtr=8.0.0.rc2_3575_g253f75b6d
+  - nextpnr-xilinx
   - pip
   - pip:                            # Packages installed from PyPI
     - -r file:requirements.txt
--- a/exhaust.py
+++ b/exhaust.py
@@ -7,6 +7,7 @@ import multiprocessing as mp
 import progressbar as pb
 import time
 import re
+import tqdm
 from contextlib import redirect_stdout
 from terminaltables import SingleTable
 from colorclass import Color
@@ -206,14 +207,10 @@ def main():
 
     assert len(tasks), "No tasks to run!"
 
-    jobs = mp.Pool(mp.cpu_count()).map_async(worker, tasks)
-    widget = ['Exhaust progress: ', pb.SimpleProgress(), pb.Bar()]
-    bar = pb.ProgressBar(widgets=widget, maxval=len(tasks)).start()
-
-    while not jobs.ready():
-        bar.update(len(tasks) - jobs._number_left)
-        time.sleep(1)
-    bar.update(len(tasks))
+    with mp.Pool(mp.cpu_count()) as pool:
+        for _ in tqdm.tqdm(pool.imap_unordered(worker, tasks),
+                           total=len(tasks), unit='test'):
+            pass
 
     # Combine results of all tests
     print('Merging results')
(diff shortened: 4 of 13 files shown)


.gitmodules
```diff
@@ -25,3 +25,6 @@
 [submodule "third_party/yosys-plugins"]
 	path = third_party/yosys-plugins
 	url = https://github.com/SymbiFlow/yosys-symbiflow-plugins.git
+[submodule "third_party/nextpnr"]
```
mithro, line 28: FYI -- We should add this to conda too.

Makefile
```diff
@@ -24,6 +26,15 @@ build-tools:
 	cd third_party/yosys && PREFIX=${INSTALL_DIR} $(MAKE) -j`nproc` && PREFIX=${INSTALL_DIR} $(MAKE) install
 	# Build Yosys plugins
 	cd third_party/yosys-plugins && export PATH=${INSTALL_DIR}/bin:${PATH} && $(MAKE)
```
mithro, line 28: The `build-tools` target should probably be split into multiple `build-tools-XXX` targets.

Makefile
```diff
@@ -24,6 +26,15 @@ build-tools:
...
+	# Build nextpnr
+	cd third_party/nextpnr && cmake -DARCH=xilinx . && make -j`nproc`; \
+		cp nextpnr-xilinx ${INSTALL_DIR}/bin; \
+		NEXTPNR_DIR=${INSTALL_DIR}/share/nextpnr; \
+		mkdir -p $$NEXTPNR_DIR; \
```
mithro, line 33: It seems a bit weird that you have to run these commands for nextpnr?

exhaust.py
```diff
@@ -18,6 +18,7 @@
 MANDATORY_CONSTRAINTS = {
     "vivado": "xdc",
     "vpr": "pcf",
+    "nextpnr": "xdc",
```
mithro, line 21: Can nextpnr read xdc now? I thought you have to specify constraints using a Python language thingy?

src/blinky/constr/nextpnr/xc7_a35t_csg324-1_arty.xdc
```diff
@@ -0,0 +1,51 @@
+# R
```
mithro, line 1: Shouldn't the constraints be the same for nextpnr/vivado/vpr?

symbiflow.py
```diff
@@ -154,22 +157,27 @@ def max_freq(self):
...
+                    )
 
         for clk in freqs:
             criticals = self.get_critical_paths(clk, 'setup')
             clocks[clk] = dict()
             clocks[clk]['actual'] = freqs[clk]
             if criticals is not None:
-                clocks[clk]['requested'] = 1e9 / criticals[clk]['requested']
+                clocks[clk]['requested'] = safe_division_by_zero(
+                    1e9, criticals[clk]['requested']
```
mithro, line 170: Why is requested zero?

symbiflow.py
```diff
@@ -321,3 +329,158 @@ def check_env():
             'yosys': have_exec('yosys'),
             'vpr': have_exec(VPR.vpr_bin()),
         }
+
+
+class Nextpnr(Toolchain):
```
mithro, line 334: Should this be NextpnrXilinx or something? It wouldn't work for iCE40 or ECP5 from what I can see here?

Makefile
acomodi: Done, each tool is under `build-<toolname>`

Makefile
acomodi: I didn't notice I could use the INSTALL PREFIX here as well, removed the `cp` of the binary.

For what regards the other files (`bba` and `bin`) they need to be generated somewhere, and there should be one for each supported device. I think they are the equivalent of the `rr_graph`. Anyways, their generation could be moved in some other make target maybe, but they definetly need to be build before tests run.

exhaust.py
acomodi: As far as I have seen in the nextpnr-xilinx examples, xdc files are used to constraint IO pads

src/blinky/constr/nextpnr/xc7_a35t_csg324-1_arty.xdc
acomodi: So, vivado and nextpnr should be able to share the xdc constraints. Cannot be the same for the vpr symbiflow toolchain.

I think that a work to be done is to be able to express all the constraints (clock, IOs) in the xdc format for the symbiflow toolchain as well, so that we can have a unified constraint file for all the toolchains.

As it is currently, the tests are being run by the exhaust based on the presence of a mandatory constraint file in the srcs directory.

symbiflow.py
acomodi: Didn't analyze this thoroughly yet.

I think the issue here is that VPR reports differently the timing information based on whether the design is single- or multi-clocked. I htink it is a matter of correctly parsing the results.

conf/environment.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
   - yosys
   - yosys-plugins
   - vtr=8.0.0.rc2_3575_g253f75b6d
+  - nextpnr-xilinx
   - pip
   - pip:                            # Packages installed from PyPI
     - -r file:requirements.txt
```
mithro, line 14: We probably want an `conf/environment.yml` and a `conf/environment.pinned.yml` file which has specific versions of yosys / vtr / nextpnr / etc. 

Can you log an issue about that?

src/blinky/constr/nextpnr/xc7_a35t_csg324-1_arty.xdc
mithro: Please log an issue about making everyone use the same constraint file.

PR #58, opened 2020-04-11
Add kokoro configuration

mithro:
Simple presubmit and continuous configurations.


mithro:
Seems to be wirking - https://source.cloud.google.com/results/invocations/f632c3ea-eb37-4e8b-823c-665166372004/targets/foss-fpga-tools%2Ffpga-tool-perf%2Fpresubmit/log

PR #62, opened 2020-04-14
exhaust: better build dir extraction and added user selection options

acomodi:

This PR enables the user to select family/device/package/board when running exhaust.

### DIFF
--- a/exhaust.py
+++ b/exhaust.py
@@ -36,9 +36,13 @@ def get_reports(out_prefix):
 
 def get_builds(out_prefix):
     """Returns all the paths of all the builds in the build directory."""
-    return matching_pattern(
-        os.path.join(root_dir, out_prefix, '*/'), '.*\/(.*)\/$'
-    )
+    build_dir = os.path.join(root_dir, out_prefix)
+    builds = []
+    for content in os.listdir(build_dir):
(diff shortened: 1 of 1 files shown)


exhaust.py
```diff
@@ -150,7 +161,8 @@ def main():
         description='Exhaustively try project-toolchain combinations'
     )
     parser.add_argument('--family', default=None, help='device family')
-    parser.add_argument('--part', default=None, help='device part')
+    parser.add_argument('--device', default=None, help='FPGA device')
```
mithro, line 164: Please add some examples to the help.

mithro:
Merge if @kgugala approves.

exhaust.py
acomodi: Done

acomodi:
There is a problem that, in CI, the result table that is printed does not contain all the tests that have been run.

Could be to the fact that all of them do fail.

@mithro Is Vivado available in CI, if so, the path is always the same? (e.g. `/opt/Xilinx/Vivado/...`)

mithro:
@acomodi - I think it's reasonable to expect Vivado to always be installed in the same place.

acomodi:
Green CI, merging.

PR #68, opened 2020-04-17
Add conda packages

acomodi:
This PR adds the conda packages and avoids the need of bulding the tools.


.github/kokoro/test.sh
```diff
@@ -13,6 +13,7 @@ echo "==========================================="
 echo "Running FPGA perf tool tests (make run-all)"
 echo "-------------------------------------------"
 (
+    source env.sh
```
mithro, line 16: Weird indenting.

Makefile
```diff
@@ -5,28 +5,17 @@ INSTALL_DIR = ${PWD}/third_party/install
...
 	git submodule update --init --recursive
-	virtualenv --python=python3 env
-	# Install fasm from third_party
-	$(IN_ENV) pip install --upgrade -e third_party/fasm
-	# Install edalize from third_party
-	$(IN_ENV) pip install --upgrade -e third_party/edalize
+	mkdir -p env
+	source utils/conda.sh
 	# Install requirements
 	$(IN_ENV) pip install -r requirements.txt
```
mithro, line 14: We should use an `environment.yaml` file to create the environment. See how sv-tests does it.

env.sh
```diff
@@ -1,14 +1,20 @@
...
 
-export FPGA_TOOL_PERF_DIR=`pwd`
+. ./env/conda/bin/activate
+source utils/environment.python.sh
 
 if [ -z "${VIVADO_SETTINGS}" ]; then
-    echo "ERROR: \$VIVADO_SETTINGS environmental variable needs to be set"
-    return 1
+    echo "WARNING: using default vivado settings"
+    VIVADO_SETTINGS=/opt/Xilinx/Vivado/2017.2/settings64.sh
```
mithro, line 8: Can you add a `FIXME:` to use the conda `xilinx-vivado` virtual package when @PiotrZierhoffer's team creates that.

env.sh
```diff
@@ -1,14 +1,20 @@
...
 fi
 
 source ${VIVADO_SETTINGS}
```
mithro, line 11: I'm pretty sure you don't want to source the vivado settings here. You should source then only *just* before you run a vivado command otherwise they mangle things like your libc.

utils/conda.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
mithro, line 1: Let's convert to using environment.yml files in the same way that sv-tests does.

utils/environment.python.sh
```diff
@@ -1,5 +1,5 @@
 # FIXME: fasm should be installed into the running Python environment.
-BASE_DIR=${FPGA_TOOL_PERF_DIR}
+BASE_DIR=$(pwd)
```
mithro, line 2: This seems wrong as it sets it to the current working directory rather than a real path?

utils/conda.sh
acomodi: As far as I understand the environment.yml is used to add the packages (e.g. yosys, vtr, etc.). The conda.sh script is used to download conda and add a basic config. I believe this is still needed

.github/kokoro/test.sh
acomodi: Done

utils/conda.sh
mithro: Yes, conda.sh still needs to exist, it should just use environment.yml as you have done now.

env.sh
acomodi: This requires changes in edalize, so that each time a vivado command is called, if the setting options is enabled. Changes are on a PR here: https://github.com/SymbiFlow/edalize/pull/31

utils/environment.python.sh
acomodi: now the path is stored in an env variable when sourcing env.sh

acomodi:
@mithro CI went green. Currently only the vivado-yosys tests are enabled. IMO this could be merged if there are no other changes to be made.

PR #78, opened 2020-04-24
Add daveshah1 nes up5k demo

### LINKED ISSUE #18: Add daveshah1's nes up5k-demo
@daveshah1 has some pretty nice up5k demos at https://github.com/daveshah1/up5k-demos/tree/master

As we really care about the up5k, it seems like a good idea to add these demos to the perf tool.

wtatarski:
Fixes #18


kamilrakoczy:
This PR is using custom ``MicroCode.v`` with removed underscores (``_``) if it is a first character in the number as this is not allowed in verilog standard.

I created PR to fix this upstream: https://github.com/daveshah1/up5k-demos/pull/2.

project/up5k-nes.json
```diff
@@ -0,0 +1,34 @@
+{
+    "srcs": [
+	"src/up5k-nes/MicroCode.v",
```
acomodi, line 3: I think this should end up in the third_party directory as well. Given that up5k demo is a submodule, you may want to create a third_party dir like `up5k-demos-changes`, where you put the MicroCode design with the fixes for the initial `_` values.

This should be similar to [ibex](https://github.com/SymbiFlow/fpga-tool-perf/tree/master/third_party/ibex) for instance, where there are, the modified files, a `README.fpga-tool-perf` file indicating the changes made, and the up5k-demo LICENSE file.

acomodi (changes requested):
Comment in

project/up5k-nes.json
kamilrakoczy: Ok, I moved this file to third party dir and added both ``README.fpga-tool-perf`` and ``LICENSE`` files.

acomodi (approved):
LGTM

PR #89, opened 2020-04-28
Use new symbiflow

acomodi:
This PR changes the symbiflow package that has new files location, as well as a far more reduced size.

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -16,5 +16,7 @@ if [ -z "${SYMBIFLOW}" ]; then
 fi
 
 export PATH=${SYMBIFLOW}/bin:${PATH}
+export PYTHONPATH=${FPGA_TOOL_PERF_BASE_DIR}/third_party/prjxray:${PYTHONPATH}
 export XRAY_DATABASE_DIR=${FPGA_TOOL_PERF_BASE_DIR}/third_party/prjxray-db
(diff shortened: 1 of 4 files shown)


env.sh
```diff
@@ -16,5 +16,7 @@ if [ -z "${SYMBIFLOW}" ]; then
 fi
 
 export PATH=${SYMBIFLOW}/bin:${PATH}
+export PYTHONPATH=${FPGA_TOOL_PERF_BASE_DIR}/third_party/prjxray:${PYTHONPATH}
```
mithro, line 19: Why is the `PYTHONPATH` back?

acomodi: I actually need to check whether prjxray can be added in the python requirements, if that is possible I will use that.

acomodi: For now this should be like this unfortunately. We could create yet another python package to extract the various prjxray scripts and install that from the requirements. Currently, the pythonpath needs to be there. 

I should add a FIXME though and a GH issue.

mithro: I'm always more happy for you to land chances with FIXME's if they have a GitHub issue number associated with them.

acomodi: Actually I have fixed this in https://github.com/SymbiFlow/prjxray/pull/1316. When that gets merged this PR should go green without the PYTHONPATH fix

acomodi: @mithro PYTHONPATH fixed, there is no more need for that anymore

PR #99, opened 2020-05-04
Create dataframe

acomodi:
This PR add the possiblity to create a pandas dataframe that can be queried to extract data of the various runs (can be adopted in colab)


acomodi:
@mithro I am also creating a colab page with a set of scripts and utility functions to easily navigate the dataframe.

PR #101, opened 2020-05-05
Add colab utils scripts

acomodi:
This PR adds some utility scripts that are useful to use when accessing data in the DataFrame, especially when using colab to visualize and exctract information.


acomodi:
@mithro FYI, the initial colab page is here: https://colab.research.google.com/drive/1ShBE2FrXEh33LD01e6hh4yM0H9p4sHYM

It uses the DataFrame class added with this PR to easily gather data.
By enhancing the class, more detailed reports can be extracted and visualized

acomodi:
@litghost This is a utility class. It gets the dataframe produced here https://github.com/SymbiFlow/fpga-tool-perf/blob/85e3d7e0e3280e4e2a7e68e0c8312b6c0b305189/exhaust.py#L262-L263

I have opted to get a utility here and clone tool-perf within colab so that colab will have code that only queries the DataFrame class to get and visualize data

HackerFoo:
Why are these helper functions necessary? They seem like accessors, but DataFrames already have these automatically.

It's best to avoid iteration and perform operations on arrays (NumPy / Matlab style.)

acomodi:
@HackerFoo So, one of the main reasons I made these accessors utilities is only due to the fact that I was dealing with non-homogeneous data.
For instance, the runtimes extracted are currently different between the different tools. 
E.g. VPR has `(synth, pack, place, route, bitstream)`, while Vivado has `(synthesis, opt_design, place_design, route_design)`.

I believe that the path forward is actually to first generate and homogenize data in a standard structure (that for instance will have blank spaces relative to opt_design when the tool is VPR).

Once the structure is standardized, the helper functions won't be needed anymore I suppose, and data can be accessed directly in colab.

acomodi:
This is not required anymore, closing

PR #103, opened 2020-05-06
Uniform results data

acomodi:
This PR makes results more uniform and creates a true DataFrame that can be queried to extract information on the runs.

### DIFF
--- a/dataframe.py
+++ b/dataframe.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+
+import pandas
+
+
+def get_clock_dataframe(results):
+    designs = []
+
+    # Clock-related lists
+    actual_frequency = []
+    requested_frequency = []
+    hold_violation = []
+    setup_violation = []
+    clock_met = []
+    clock_names = []
+
+    for idx, design in enumerate(results['design']):
+        clock = results['max_freq'][idx]
+        if clock:
+            for key, value in clock.items():
+                designs.append(design)
+                clock_names.append(key)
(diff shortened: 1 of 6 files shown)


acomodi:
@HackerFoo @litghost @mithro  FYI. [This is the colab page](https://colab.research.google.com/drive/1ShBE2FrXEh33LD01e6hh4yM0H9p4sHYM?authuser=1#scrollTo=jk_IBrHiop-A&uniqifier=3) that uses the dataframe produced with this PR.

dataframe.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
kgugala, line 1: License header is missing here

symbiflow.py
```diff
@@ -443,13 +447,13 @@ def resources(self):
...
-            "GLB": "unsupported",
-            "PLL": "NA",  #str(pll),
-            "IOB": "NA",  #str(iob),
+            "LUT": None,  #str(lut),
+            "DFF": None,  #str(dff),
+            "BRAM": None,  #str(bram),
+            "CARRY": None,  #str(carry),
+            "GLB": None,
+            "PLL": None,  #str(pll),
+            "IOB": None,  #str(iob),
```
kgugala, line 456: can we remove those comments?

dataframe.py
acomodi: Done

symbiflow.py
acomodi: Done

PR #112, opened 2020-05-15
Custom parameter injection and multiple tests run

acomodi:
This PR has a double purpose:

- allowing to inject custom parameters for the tool (currently it is enabled for VPR only, but could be expanded in the future to be more flexible).
- make it possible to easily running multiple times the tests to gather more samples.

Note: Given that the exhaust script is used mainly to perform a batch of tests, I have limited the freedom degree to be able to choose only `toolchains` and `projects` (from a user perspective).
The user can still run single tests using `fpgaperf.py`.

TODO:
- [x] test multiple runs with all of them gathering information on CI
- [x] test multiple runs with different options for VPR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,13 @@ conda:
 	cd third_party/prjxray && $(MAKE) build -j`nproc`
 
 run-all:
-	python3 exhaust.py --fail
+	python3 exhaust.py
+	# TMP: Testing multiple options
+	python3 exhaust.py --parameters parameters.json --toolchain vpr --build_type parameters --project blinky
+	# TMP: Testing multiple builds
+	for run in {1..10}; do \
+		python3 exhaust.py --toolchain vpr --project blinky --build_type multiple-samples --build $$run; \
+	done
 
 PYTHON_SRCS=$(shell find . -name "*py" -not -path "./third_party/*" -not -path "./env/*")
 
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ conda:
 	cd third_party/prjxray && $(MAKE) build -j`nproc`
 
 run-tests:
-	python3 exhaust.py --build_type generic
+	python3 exhaust.py --build_type generic --fail
 
 run-parameters-tests:
 	python3 exhaust.py --parameters parameters.json --toolchain vpr --build_type parameters
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ conda:
 	cd third_party/prjxray && $(MAKE) build -j`nproc`
 
 run-all:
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -18,8 +18,21 @@ conda:
...
 	cd third_party/prjxray && $(MAKE) build -j`nproc`
 
+run-tests:
+	python3 exhaust.py --build_type generic --fail
+
+run-parameters-tests:
+	python3 exhaust.py --parameters parameters.json --toolchain vpr --build_type parameters
+
+run-multiple-samples-tests:
+	for run in {0..10}; do \
```
kgugala, line 28: can we define this number as parameter? Preferably sth we can overwrite by setting an ENV variable. The dafault can be 10

acomodi: Done, set this to 2 as default actually, as this CI should just test that the behavior is correct.

With another script we can make use of the multiple test runs to have a continues build that is triggered on every HEAD change.

acomodi:
@kgugala I have updated the [colab page](https://colab.research.google.com/drive/1ShBE2FrXEh33LD01e6hh4yM0H9p4sHYM) with the new results

kgugala:
> @kgugala I have updated the [colab page](https://colab.research.google.com/drive/1ShBE2FrXEh33LD01e6hh4yM0H9p4sHYM) with the new results
Those results are identical (up to 7th digit after comma). Are you sure this is correct?

Screenshot from 2020-05-19 14-03-20

kgugala:
Screenshot from 2020-05-19 14-06-39

What does this `None` parameters mean? I don't think VPR was ran there w/o any parameters

acomodi:
> What does this `None` parameters mean? I don't think VPR was ran there w/o any parameters

Actually this is a bug I discovered just now, the `parameters` run had an additional build with no parameters override set, hence the `None` value. I'll push a fix in a moment


> Those results are identical (up to 7th digit after comma). Are you sure this is correct?

Looking at the meta.json that is produced by each run it seems that those are indeed identical. I think this is good news actually as this seems to indicate that, with identical input conditions we get identical outputs.

acomodi:
In addition, there are some fixes still to be made:

- [ ] when using the parameters build type, the dataframe should contain columns with the parameters values, so that they can be correctly displayed in a more readable format.
- [ ] fixes to results parsing, as, by looking at the table, there seem to be some issues related to that (e.g. not all clocks are correctly gathered, missing placing time for vivado, resource usage in VPR is way to less than vivado, etc.) 

I believe that, if everything is ok with this PR we should get to merge it as it is quite big, and move with the fixes in a second step.

PR #124, opened 2020-05-26
Add fasm2bels

acomodi:
This PR adds fasm2bels support.

To remove the WIP, the following PRs must be merged:
- https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/2: this enables to use the SymbiFlow python package
- https://github.com/SymbiFlow/symbiflow-rr-graph/pull/1: same as above
- https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1504: This allows to not rely on having prjxray submodule


src/baselitex/constr/vpr-fasm2bels/xc7_a35t_csg324-1_arty.pcf
```diff
@@ -0,0 +1,86 @@
+# 100 MHz CLK
```
kgugala, line 1: why do we need a copy of the constraints? I'd expect the fasm2bels flow will reuse the data from original design.

acomodi: Given that vpr-fasm2bels is a different toolchain, it currently requires to have its own constr directory.
I think that yet another refactor could be made to have a better handling of these constraints, so to avoid having to duplicate them, I'll open an issue about that.

Furthermore, `vpr` and `vpr-fasm2bels` builds of the same design need to be in two distinct directories, otherwise they would conflict on the generated files when running the exhaust script. This prevents us from reusing the generated files for the fasm2bels toolchain.

kgugala: sure, but the whole idea of having `fasm2bels` here is to compare the timing info we get from VPR with vendor tools. To do that we have to run both tools on the same design. Making a copy of the constraint files makes it hard to maintain.

acomodi: Indeed, there is a plan [here](https://github.com/SymbiFlow/fpga-tool-perf/issues/125#issuecomment-633981849) to fix this. But I think that this is something for a followup PR

acomodi: @kgugala FYI, I have changed the way constraints are applied in [this commit](https://github.com/SymbiFlow/fpga-tool-perf/pull/124/commits/670c120f9cd9ece1762f14e849c8e7268384c9ab). This way we can avoid duplicating constr files that should be the same for different toolchains.

PR #136, opened 2020-05-29
Add nextpnr ice40 support

acomodi:
This PR requires the new nextpnr-ice40 conda package from https://github.com/SymbiFlow/conda-packages/pull/96 as it cannot currently be installed due to having a wrong python dependency.


conf/requirements.txt
```diff
@@ -9,7 +9,7 @@ intervaltree
 jinja2
 pytest
 pandas
--e git://github.com/SymbiFlow/fasm.git#egg=fasm
--e git://github.com/SymbiFlow/prjxray.git#egg=prjxray
--e git://github.com/SymbiFlow/edalize.git@symbiflow#egg=edalize
--e git://github.com/antmicro/symbiflow-xc-fasm2bels.git@add-fasm2bels#egg=symbiflow-xc-fasm2bels
+git+git://github.com/SymbiFlow/fasm.git#egg=fasm
```
mithro, line 12: This most certainly looks wrong?

acomodi: I made this change mainly to avoid having the various repos to be cloned locally. In this way, they get cloned in a tmp dir, installed in the environment and deleted

PR #143, opened 2020-06-09
Add the attosoc test from nextpnr-xilinx

HackerFoo:


project/attosoc.json
```diff
@@ -0,0 +1,20 @@
...
+    ],
+    "top": "top",
+    "name": "attosoc",
+    "toolchains": {
+        "vivado": {
+            "arty": ["arty.xdc"]
+        },
+        "nextpnr": {
+            "arty": ["arty.xdc"]
+        }
```
litghost, line 17: Should we have an issue to try adding vpr here once IO LOC support for symbiflow-arch-defs is added (e.g. https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1519 )

acomodi: I believe we can temporarily use PCF and convert it to XDC once the support is ready.

On a side note, with https://github.com/SymbiFlow/fpga-tool-perf/pull/136, nextpnr toolchain moved to be `nextpnr-xilinx` as I have added ice40 support (which resulted in the need renaming nextpnr toolchains).

project/picorv32.json
```diff
@@ -13,6 +13,9 @@
         "vpr": {
             "arty": ["arty.pcf"],
             "zybo": ["zybo.pcf"]
+        },
+        "nextpnr": {
```
acomodi, line 17: `nextpnr-xilinx`

acomodi:
I believe the failure in CI is related to an outdated version of Yosys. As soon as the master+wip branch gets updated, this should turn green.

mithro:
@acomodi - We should resurrect or close out this pull request.

PR #151, opened 2020-06-16
--verbose functionality in fpgaperf.py and exhaust.py

ryancj14:
This pull request does two things.

First, it adds --verbose functionality to `fpgaperf.py` and `exhaust.py`. This will allow users and developers to more easily debug issues and only display build stats in the terminal if they want it.

An example run of `fpgaperf.py` **without --verbose** would look like this:
```
$ python3 fpgaperf.py --project oneblink --toolchain vpr --board arty
Writing to build/oneblink_vpr_xc7_a35tcsg324-1_arty_pcf_carry-n
$
```

An example run of `fpgaperf.py` **with --verbose** would look like this:
```
$ python3 fpgaperf.py --project oneblink --toolchain vpr --board arty --verbose
Parsing Arguments......

Continuing.............

Getting Constraints....

Starting Project.......
Writing to build/oneblink_vpr_xc7_a35tcsg324-1_arty_pcf_carry-n

Running Project........

Printing Stats.........

===============================
Setting
===============================

+-----------+------------------------------------------------+
| Settings  | Value                                          |
+-----------+------------------------------------------------+
| Design    | oneblink_vpr_xc7_a35tcsg324-1_arty_pcf_carry-n |
| Family    | xc7                                            |
| Device    | a35t                                           |
| Package   | csg324-1                                       |
| Project   | oneblink                                       |
| Toolchain | vpr                                            |
| Strategy  | None                                           |
| Carry     | False                                          |
| Seed      | default                                        |
+-----------+------------------------------------------------+

===============================
Clocks
===============================

+--------------+-------------+----------------+---------+-----------------+----------------+
| Clock domain | Actual freq | Requested freq | Met?    | Setup violation | Hold violation |
+--------------+-------------+----------------+---------+-----------------+----------------+
| clk          | 185.247 MHz | 0.000 MHz      | unknown | 0.000 ns        | 0.000 ns       |
+--------------+-------------+----------------+---------+-----------------+----------------+

===============================
Toolchain Run-Times
===============================

+--------------+--------------------+
| Stage        | Run Time (seconds) |
+--------------+--------------------+
| prepare      | 0.139              |
| synthesis    | 0.013              |
| optimization | N/A                |
| packing      | 0.004              |
| placement    | 0.009              |
| routing      | 0.004              |
| fasm         | 0.004              |
| checkpoint   | N/A                |
| bitstream    | 0.004              |
| reports      | N/A                |
| total        | 0.176              |
| nop          | 0.002              |
| fasm2bels    | N/A                |
+--------------+--------------------+

===============================
FPGA resource utilization
===============================

+----------+------+
| Resource | Used |
+----------+------+
| BRAM     | 0    |
| CARRY    | 6    |
| DFF      | 24   |
| GLB      | N/A  |
| IOB      | 2    |
| LUT      | 1    |
| PLL      | 0    |
+----------+------+

Writing Metadata.......

$
```

The second thing this pull request does is change the default location to which `exhaust.py` saves its builds. The old default was the `build` directory, where it gets thrown in with the `fpgaperf.py` runs and causes exhuast.py to fail in the print_summary_table() method. The new default is the `build/_exhaust-runs` subdirectory, where it is still in the `build` directory and is easy to find, but separated from the other builds.


mithro:
Hi,

Thank you for sending the pull request! It's awesome to see you contributing to this project. However, before I can look at this request you need to add [DCO sign-off](https://developercertificate.org/) to commits in this repository.

We require [DCO sign-off](https://developercertificate.org/) for **all** commits to this repository (including all authors). Please see the [following URL](https://github.com/probot/dco#how-it-works) to see more information and BeeWare also have a [Beginners guide to DCOs](https://pybee.org/contributing/how/dco/).

You can use `git commit --amend --signoff` and then a `git push --force` to add the signoff. You can also use `git rebase --signoff`.

Thanks!

ryancj14:
Ok Thanks! Will do!

exhaust.py
```diff
@@ -116,10 +116,16 @@ def main():
 
     args = parser.parse_args()
 
+    if args.verbose:
```
acomodi, line 119: Instead of having all these `if args.verbose` it may be better to make use of the python `logging` library.
With these we can set a logger that add logs based on a logging level (in this case the logging level would be `DEBUG`).

fpgaperf.py
```diff
@@ -200,8 +204,13 @@ def run(
         out_prefix=out_prefix,
     )
 
+    if verbose:
+        print("\nRunning Project........")
     t.run()
-    print_stats(t)
+    if verbose:
+        print("\nPrinting Stats.........")
+        print_stats(t)
```
acomodi, line 212: `print_stats` should be run also in the non-verbose mode

exhaust.py
ryancj14: Sorry it took me so long to get this done. I've implemented the logging. Is this how you would like it? Also, I put the logging level as INFO (forgetting that you had mentioned DEBUG as the logging level). Would you like me to change it to DEBUG?

acomodi: Hi, no problem.
Regarding the logging info I would say it would be better to keep them as debug

acomodi:
@ryancj14 I believe you have commited also generated files from the build. Also there is an issue with the DCO still that should be solved as well (https://github.com/SymbiFlow/fpga-tool-perf/pull/151#issuecomment-644947446).

ryancj14:
Oh yeah, I see that DCO issue. However, upon running the command it gives me to fix it, it says there is no signoff option. I tried changing it to -S but that didn't work either. I've learned a lot from this PR. Honestly, it might be a good idea to just delete this PR and have me fork the repo again and be careful to sign off and only add the two files to the commit.

PR #176, opened 2020-07-02
Implementing --verbose functionality using logger

ryancj14:
Added a logger for both `fpgaperf.py` and `exhaust.py`, that is turned on if `--verbose` is specified. Also added messages that will be printed through the logger under the `DEBUG` level.

```
student@ubuntu:~/fpga-tool-perf$ python3 fpgaperf.py --verbose --project oneblink --toolchain vivado --board basys3
DEBUG: Parsing Arguments
DEBUG: Continuing
DEBUG: Getting Constraints
DEBUG: Starting Project
Writing to build/oneblink_vivado_xc7_a35tcpg236-1_basys3_xdc_carry-n
DEBUG: Running Project
DEBUG: Printing Stats

===============================
Setting
===============================

+-----------+-----------------------------------------------------+
| Settings  | Value                                               |
+-----------+-----------------------------------------------------+
| Design    | oneblink_vivado_xc7_a35tcpg236-1_basys3_xdc_carry-n |
| Family    | xc7                                                 |
| Device    | a35t                                                |
| Package   | cpg236-1                                            |
| Project   | oneblink                                            |
| Toolchain | vivado                                              |
| Strategy  | None                                                |
| Carry     | False                                               |
| Seed      | default                                             |
+-----------+-----------------------------------------------------+

===============================
Clocks
===============================

+--------------+-------------+----------------+------+-----------------+----------------+
| Clock domain | Actual freq | Requested freq | Met? | Setup violation | Hold violation |
+--------------+-------------+----------------+------+-----------------+----------------+

===============================
Toolchain Run-Times
===============================

+--------------+--------------------+
| Stage        | Run Time (seconds) |
+--------------+--------------------+
| prepare      | 0.487              |
| synthesis    | 37.000             |
| optimization | 19.000             |
| packing      | N/A                |
| placement    | N/A                |
| routing      | 19.000             |
| fasm         | N/A                |
| checkpoint   | N/A                |
| bitstream    | 10.000             |
| reports      | N/A                |
| total        | 214.221            |
| nop          | 0.003              |
| fasm2bels    | N/A                |
+--------------+--------------------+

===============================
FPGA resource utilization
===============================

+----------+------+
| Resource | Used |
+----------+------+
| BRAM     | 0    |
| CARRY    | 6    |
| DFF      | 24   |
| GLB      | N/A  |
| IOB      | 2    |
| LUT      | 1    |
| PLL      | 0    |
+----------+------+
DEBUG: Writing Metadata
Unable to recognize Vivado version
```

Also changed the default directory where `exhaust.py` will save its builds (to a subdirectory of build: `build/_exhaust-runs`) for a better chance that there will be no errors while it is printing out stats on the builds.

### DIFF
--- a/fasm2bels.py
+++ b/fasm2bels.py
@@ -37,16 +37,11 @@ class VPRFasm2Bels(VPR):
         return get_vivado_max_freq(report_file)
 
     def run_steps(self):
-        with Timed(self, 'synthesis'):
-            self.backend.build_main(self.top + '.eblif')
-        with Timed(self, 'pack'):
-            self.backend.build_main(self.top + '.net')
-        with Timed(self, 'place'):
-            self.backend.build_main(self.top + '.place')
-        with Timed(self, 'route'):
-            self.backend.build_main(self.top + '.route')
-        with Timed(self, 'fasm'):
-            self.backend.build_main(self.top + '.fasm')
+        self.backend.build_main(self.top + '.eblif')
+        self.backend.build_main(self.top + '.net')
+        self.backend.build_main(self.top + '.place')
+        self.backend.build_main(self.top + '.route')
+        self.backend.build_main(self.top + '.fasm')
         with Timed(self, 'bitstream'):
             self.backend.build_main(self.top + '.bit')
         with Timed(self, 'fasm2bels'):
--- a/symbiflow.py
+++ b/symbiflow.py
@@ -21,16 +21,11 @@ class VPR(Toolchain):
         self.dbroot = None
 
     def run_steps(self):
-        with Timed(self, 'synthesis'):
-            self.backend.build_main(self.top + '.eblif')
-        with Timed(self, 'pack'):
-            self.backend.build_main(self.top + '.net')
-        with Timed(self, 'place'):
-            self.backend.build_main(self.top + '.place')
-        with Timed(self, 'route'):
-            self.backend.build_main(self.top + '.route')
-        with Timed(self, 'fasm'):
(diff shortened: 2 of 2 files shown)


fpgaperf.py
```diff
@@ -382,6 +403,7 @@ def main():
                 print('{}: error: {}'.format(sys.argv[0], e))
             sys.exit(1)
 
+        logger.debug("Continuing")
```
acomodi, line 406: Probably this is not useful as it does not give information on what is happening

ryancj14: Good point. `Preparing Project` would probably be a better descriptor for what is happening there. I'll change that.

acomodi: Hmm maybe you could actually skip that entirely as I already see there is something similar (`Starting Project`). Might be confusing

ryancj14: My thought process was that first, it prepares the project (with get_project(project) and specifying the board info and toolchain info that it would need) and then (after `Getting Constraints`) it starts the project with t.project() and runs the project with t.run(). Would you still like me to get rid of the Preparing Project? Another option could be to remove the `Starting Project` and move `Running Project` to be before t.project(). What do you think?

acomodi: I believe the second option is all right (replace starting project with running project)

acomodi: One more thing is that you probably should run `make format` to have the autoformatter clean the python changes and conform them to the standard used for this project

PR #182, opened 2020-07-20
fasm2bels: add capnp schema dir

acomodi:


litghost:
Shouldn't this PR use a new VPR package?

acomodi:
> Shouldn't this PR use a new VPR package?

Yes. Actually, It already does, I have pinned just the VPR version, and that was untouched in the conda-packages update

PR #186, opened 2020-07-22
Using Logger to implement --verbose functionality

ryancj14:
New PR with identical changes as in [PR #176](https://github.com/SymbiFlow/fpga-tool-perf/pull/176).

Added a logger for both fpgaperf.py and exhaust.py, that is turned on if --verbose is specified. Also added messages that will be printed through the logger under the DEBUG level.

Also changed the directory to which exhaust runs are saved from `build` to `build/_exhuaust-runs`.

### DIFF
--- a/exhaust.py
+++ b/exhaust.py
@@ -5,6 +5,7 @@ import os
 import glob
 import time
 import re
+import logging
 from terminaltables import AsciiTable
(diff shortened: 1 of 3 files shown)


acomodi:
Thanks for the enhancement

PR #187, opened 2020-07-23
Attosoc

acomodi:
This builds on top of https://github.com/SymbiFlow/fpga-tool-perf/pull/143


acomodi:
> It would be good to also explain the change to `murax.json`.

This is for fasm2bels to assign the input clock, so that the report contains the correct CPD information.

mithro:
@acomodi - We should resurrect or close out this pull request.

PR #201, opened 2020-08-03
Add overhead to VPR run-times

### LINKED ISSUE #195: Add VPR circuit cleaning step run-time
There should be another entry in the run-times called "overhead" which collects the time the tools do not directly do synthesis or P&R.

kamilrakoczy:
Fixes #195

### DIFF
--- a/symbiflow.py
+++ b/symbiflow.py
@@ -377,6 +377,22 @@ class VPR(Toolchain):
                     if '{} took'.format(step) in l:
                         return float(l.split()[position])
 
+        def get_overhead_runtime(logfile):
+            steps = [
+                'Clean circuit took', 'Load Timing Constraints took',
+                'Create Device took'
+            ]
+            overhead = 0.0
+            with open(logfile, 'r') as fp:
+                for l in fp:
+                    l = l.strip()
+                    for step in steps:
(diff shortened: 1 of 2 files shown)


symbiflow.py
```diff
@@ -385,8 +385,17 @@ def get_step_runtime(step, logfile, position):
         fasm_log = os.path.join(self.out_dir, 'fasm.log')
 
         log['pack'] = get_step_runtime('Packing', pack_log, 3)
+        log['overhead'] = get_step_runtime(
+            'The entire flow of VPR', pack_log, 6
```
acomodi, line 389: This run-time actually includes also the time required for the actual packing. It would be better to take, for instance the following run-tmes:
```
# Clean circuit took 17.81 seconds (max_rss 1115.7 MiB, delta_rss +943.9 MiB) 
...
# Load Timing Constraints took 3.03 seconds (max_rss 1118.1 MiB, delta_rss +0.0 MiB) 
...
# Create Device took 6.76 seconds (max_rss 3472.8 MiB, delta_rss +2354.8 MiB)

```

Given that there are several of these, it would be good maybe to have a function that adds all these runtimes for each P&R step.

PR #203, opened 2020-08-03
Timing Consistency Between Toolchains

### LINKED ISSUE #221: 'nop' Runtime
I'm creating this issue just to get a discussion going about the odd 'nop' timing in the fpgperf.py run. I suggest removing it in #203. 

From what I've seen, the 'nop' timing is the same for each toolchain, simply running of the `true` command in bash, and reporting something around `0.003` or `0.004` seconds. This is the code in `toolchain.py` lines 53-54 which every toolchain inherits from:
```
with Timed(self, 'nop'):
            subprocess.check_call("true", shell=True, cwd=self.out_dir)
```
I wondered if this was an attempt to make sure that different builds being compared are running with the same processing speed, but this seems to me like it would be an ineffective way to do that. My other thought is that it may be a placeholder for something that hasn't yet been implemented but in that case, we could remove it until it is functional.

Can we get any clarification about the purpose of 'nop' in the reported runtimes? If not, can it be removed with the merge of #203?

ryancj14:
I found two issues with timing consistency that I have addressed with this pull request.

Each toolchain's runtime includes a `prepare` time, as well as a `self.add_runtimes()`. The problem is that while `prepare` time is included in symbiflow's `total` time, it is not included in vivado's total time. The other inconsistency is that the `self.add_runtimes()` function is included in symbiflow's `total` but not in vivado's.

These are my solutions:
1. In `symbiflow.py`, I removed the `self.add_runtimes()` from the `total` runtime, since it isn't actually a part of the running toolchain, simply a gathering of runtimes. This simply meant deleting a tab from it, so it takes place after (and not during) the `with Timed(self, 'total'):` code block.
2. In `vivado.py`, I moved the `with Timed(self, 'total'):` block to the front of the function, to include prepare. It looks confusing through Github, but I really just moved that one line and tabbed the `with Timed(self, 'prepare'):` block to be a part of it.

This clears up all the consistency issues that I know of.

fixes #221

### DIFF
--- a/symbiflow.py
+++ b/symbiflow.py
@@ -35,10 +35,13 @@ class VPR(Toolchain):
         self.backend.build_main(self.top + '.eblif')
         with Timed(self, 'pack_all', unprinted_runtime=True):
             self.backend.build_main(self.top + '.net')
+
         with Timed(self, 'place_all', unprinted_runtime=True):
             self.backend.build_main(self.top + '.place')
+
         with Timed(self, 'route_all', unprinted_runtime=True):
             self.backend.build_main(self.top + '.route')
+
         self.backend.build_main(self.top + '.fasm')
         with Timed(self, 'bitstream'):
             self.backend.build_main(self.top + '.bit')
@@ -159,7 +162,9 @@ class VPR(Toolchain):
                     edam=self.edam, work_root=self.out_dir
                 )
                 self.backend.configure("")
+
             self.run_steps()
+
         self.add_runtimes()
         self.add_wirelength()
         self.add_maximum_memory_use()
@@ -728,6 +733,7 @@ class NextpnrXilinx(Toolchain):
                     edam=self.edam, work_root=self.out_dir
                 )
                 self.backend.configure("")
+
             self.backend.build_main(self.project_name + '.fasm')
             self.run_steps()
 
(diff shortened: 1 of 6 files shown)


ryancj14:
The only other issue I would like to discuss in regards to this pull request is the `nop` timing.

In toolchain.py, 
this is the code that determines the runtime recorded under `nop`. It takes place in the toolchain.py `__init__` definition.
```
with Timed(self, 'nop'):
    subprocess.check_call("true", shell=True, cwd=self.out_dir)
```

To me, it seems to simply call the true command in the bash shell, and it is the same for every toolchain. Is this correct? and why is this? Is it simply a way to judge the speed of the computer you are working with before you start the toolchain? What is the importance of having this `nop` in this spot? 

An answer to this would be much appreciated. If we don't need `nop` in our timing, I can remove those lines. Thanks.

ryancj14:
This is ready to be merged.

symbiflow.py
```diff
@@ -159,11 +159,10 @@ def run(self):
                     edam=self.edam, work_root=self.out_dir
                 )
                 self.backend.configure("")
-
             self.run_steps()
-            self.add_runtimes()
-            self.add_wirelength()
-            self.add_maximum_memory_use()
+        self.add_runtimes()
```
acomodi, line 163: Let's add a new-line between different Timed scopes, just to make it clearer (same applies to the other toolchains.

acomodi (changes requested):
LGTM. apart for small changes to formatting.

acomodi (approved):
LGTM

PR #204, opened 2020-08-03
Minor Enhancements for Usability

ryancj14:
1. Adds the `Writing to <out_prefix>` statement to the exhaust run. Serves the same purpose as it does in `fpgaperf.py`.

2. Changes `os.mkdir()` to `os.makedirs()`. We recently made this change to the `exhaust.py` run. It will simply allow for more specific directory paths if the user is so inclined, without failing if there are multiple folders in the path yet to be created. This also makes it easier to use the `--out-prefix` and `--out-dir` arguments separately.

### DIFF
--- a/toolchain.py
+++ b/toolchain.py
@@ -195,7 +195,7 @@ class Toolchain:
 
         out_prefix = out_prefix or 'build'
         if not os.path.exists(out_prefix):
-            os.mkdir(out_prefix)
+            os.makedirs(out_prefix)
 
         if out_dir is None:
             out_dir = out_prefix + "/" + self.design()
@@ -203,7 +203,7 @@ class Toolchain:
         if overwrite and os.path.exists(out_dir):
             shutil.rmtree(out_dir)
         if not os.path.exists(out_dir):
-            os.mkdir(out_dir)
+            os.makedirs(out_dir)
         print('Writing to %s' % out_dir)
         data = project.get('data', None)
         if data:
--- a/toolchain.py
+++ b/toolchain.py
@@ -194,8 +194,7 @@ class Toolchain:
             self.clocks = project['clocks']
 
         out_prefix = out_prefix or 'build'
-        if not os.path.exists(out_prefix):
-            os.makedirs(out_prefix)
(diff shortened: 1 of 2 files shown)


toolchain.py
```diff
@@ -195,15 +195,15 @@ def project(
 
         out_prefix = out_prefix or 'build'
         if not os.path.exists(out_prefix):
-            os.mkdir(out_prefix)
+            os.makedirs(out_prefix)
```
kamilrakoczy, line 198: I think it would be better to remove this `if` and set `exist_ok=True` in `os.makedirs`.
It would also be good to first resolve out_prefix, as `os.makedirs` is not handling pardir (`..`) and tilde (`~`) if it is not done before.

ryancj14: Thanks for the feedback @kamilrakoczy, what do you mean by resolving `out_prefix`? As I've tested it, using `..` or `~` seems to work with `os.makedirs()`.

kamilrakoczy: Maybe it depends on python version.
Based on description in python docs: https://docs.python.org/3/library/os.html#os.makedirs
``os.makedirs()`` does not resolve ``..`` and ``~``. There is also similar question on stack overflow: https://stackoverflow.com/questions/2057045/pythons-os-makedirs-doesnt-understand-in-my-path

Doing local testing I get:
```
Python 3.6.9 (default, Jul 17 2020, 12:50:27)
>>> import os
>>> os.makedirs("~/test")
>>> os.makedirs("../test2")
grey@grey:~/python_path$ ls -la
total 20
drwxrwxr-x   3 grey grey  4096 wrz  9 08:15  .
drwxr-xr-x 112 grey grey 12288 wrz  9 08:15  ..
drwxrwxr-x   3 grey grey  4096 wrz  9 08:15 '~'
```
so ``os.makedirs("~/test")`` instead of creating a new folder in the home directory, it creates a new folder in the current directory with ``~`` as the directory name. ``os.makedirs("../test2")`` seems to work as expected, but I'm not sure if it is always the case.

ryancj14:
Ready for merge.

PR #206, opened 2020-08-04
Implemented New --list-combinations Argument

ryancj14:
This pull request fixes #202, adding a new `--list-combinations` argument, which lists all of the combinations of <project, toolchain, board> available to run in `fpga-tool-perf`. Example:
```
$ python3 fpgaperf.py --list-combinations
(Project, Toolchain, Board)
---------------------------
(baselitex, vivado, arty)
(baselitex, vivado-yosys, arty)
(baselitex, nextpnr-xilinx, arty)
(baselitex, vpr, arty)
(baselitex, vpr-fasm2bels, arty)
(blinky, vpr, arty)
(blinky, vpr-fasm2bels, arty)
(blinky, nextpnr-xilinx, arty)
(blinky, nextpnr-xilinx-fasm2bels, arty)
(blinky, vivado, arty)
(blinky, vivado-yosys, arty)
(hamsternz-hdmi, vivado, nexys)
(ibex, vpr, arty)
(ibex, vpr-fasm2bels, arty)
(ibex, vivado, arty)
(ibex, vivado-yosys, arty)
(murax, vpr, basys3)
(murax, vpr-fasm2bels, basys3)
(murax, nextpnr-xilinx, basys3)
(murax, vivado-yosys, basys3)
(murax, vivado, basys3)
(oneblink, vpr, arty)
(oneblink, vpr, basys3)
(oneblink, vpr, zybo)
(oneblink, vpr-fasm2bels, arty)
(oneblink, vpr-fasm2bels, basys3)
(oneblink, vpr-fasm2bels, zybo)
(oneblink, vivado, arty)
(oneblink, vivado, basys3)
(oneblink, vivado, zybo)
(oneblink, vivado-yosys, arty)
(oneblink, vivado-yosys, basys3)
(oneblink, vivado-yosys, zybo)
(oneblink, nextpnr-ice40, icebreaker)
(picorv32, vivado, arty)
(picorv32, vivado, zybo)
(picorv32, vivado-yosys, arty)
(picorv32, vivado-yosys, zybo)
(picorv32, vpr, arty)
(picorv32, vpr, zybo)
(picorv32, vpr-fasm2bels, arty)
(picorv32, vpr-fasm2bels, zybo)
(picosoc, vpr, basys3)
(picosoc, vpr-fasm2bels, basys3)
(picosoc, nextpnr-xilinx, basys3)
(picosoc, vivado, basys3)
(picosoc, vivado-yosys, basys3)
(picosoc-simpleuart, vivado, basys3)
[...]

PR #207, opened 2020-08-05
Exhaust Printing Fix

ryancj14:
There was an issue with `exhaust.py` runs, which would print the stats of all of the saved builds, even if only a certain project or toolchain was specifically being tested like in running `python3 exhaust.py --toolchain vivado-yosys`.

This pull request fixes this by adding conditions to only print the runs that match the specified toolchain and project arguments, just as the condition to only print runs of the specified build name and number already exists.

I also fixed the names of the `vivado-yosys` toolchain and many of the projects to match the names that the user inputs to run them. Odd to run the `vivado-yosys` toolchain and then have it called the `yosys-vivado` toolchain in the build summary. Naming the synthesis tool after the primary tool seems to match the rest of the toolchain names too. This consistency also matters for the implementation of filtering the results printed.

### DIFF
--- a/vivado.py
+++ b/vivado.py
@@ -235,7 +235,7 @@ class VivadoYosys(Vivado):
         Vivado.__init__(self, rootdir)
         self.synthtool = 'yosys'
         self.synthoptions = ['-iopad', '-arch xc7', '-flatten']
-        self.toolchain = 'vivado-yosys'
+        self.toolchain = 'yosys-vivado'
 
     @staticmethod
     def check_env():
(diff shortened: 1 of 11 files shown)


vivado.py
```diff
@@ -235,7 +235,7 @@ def __init__(self, rootdir):
         Vivado.__init__(self, rootdir)
         self.synthtool = 'yosys'
         self.synthoptions = ['-iopad', '-arch xc7', '-flatten']
-        self.toolchain = 'yosys-vivado'
+        self.toolchain = 'vivado-yosys'
```
acomodi, line 238: I think this was written as `yosys-vivado` as it better shows the fact that first yosys is run for synthesis and than vivado does P&R.

acomodi:
There is a conflict in fpgaperf.py after merging https://github.com/SymbiFlow/fpga-tool-perf/pull/206

PR #212, opened 2020-08-10
vpr: update overhead calculation

kamilrakoczy:
Update overhead calculation for vpr to `total step run` - `useful step work`.

This PR also adds possibility to calculate time with `Timed` class without automatic adding it to printable `runtimes` collection.

### DIFF
--- a/symbiflow.py
+++ b/symbiflow.py
@@ -22,9 +22,12 @@ class VPR(Toolchain):
 
     def run_steps(self):
         self.backend.build_main(self.top + '.eblif')
-        self.backend.build_main(self.top + '.net')
-        self.backend.build_main(self.top + '.place')
-        self.backend.build_main(self.top + '.route')
+        with Timed(self, 'pack_all', unprinted_runtime=True):
+            self.backend.build_main(self.top + '.net')
+        with Timed(self, 'place_all', unprinted_runtime=True):
+            self.backend.build_main(self.top + '.place')
+        with Timed(self, 'route_all', unprinted_runtime=True):
+            self.backend.build_main(self.top + '.route')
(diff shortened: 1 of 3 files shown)


symbiflow.py
```diff
@@ -401,14 +405,19 @@ def get_overhead_runtime(logfile):
...
         #      For now the whole command execution time is considered
         log['fasm'] = get_step_runtime('The entire flow of VPR', fasm_log, 6)
+        if "pack_all" in self.unprinted_runtimes and "place_all" in self.unprinted_runtimes and "route_all" in self.unprinted_runtimes and "fasm_all" in self.unprinted_runtimes:
+            log['overhead'] = self.unprinted_runtimes['pack_all'] - log['pack']
+            log['overhead'
+                ] += self.unprinted_runtimes['place_all'] - log['place']
+            log['overhead'
+                ] += self.unprinted_runtimes['route_all'] - log['route']
+            log['overhead'
+                ] += self.unprinted_runtimes['fasm_all'] - log['fasm']
```
acomodi, line 420: I would leave fasm out of these calculations, as it is required for us to modify VPR to get the useful work of the genfasm step, which is not reported in the fasm.log.

Why is the if statement needed? maybe would be worth to use the function to get the overhead time that checks if a step is in the unprinted_runtimes and, if it is, returns `pack_all - pack`

kamilrakoczy: I removed fasm from calculations and moved this code to function.

I didn't checked exactly why this if statement is needed, but generally it is failing when `add_runtimes` is called. This function is shared between `nextpnr` and  `vpr`, so I think it is failing, when `nextpnr` toolchain is selected as it do not have `pack_all` runtime.

PR #217, opened 2020-08-12
Add wirelength and max_rss to meta.json

### LINKED ISSUE #216: Add more metrics to the meta.json results files
To have a more complete view on the results obtained by the various tools, we would need to have additional metrics extracted from the log files (especially for VPR):

- Wirelength: This is found in the route.log file (VPR)
- Maximum memory use: this is found in each log file (VPR).

kamilrakoczy:
This PR assumes that `MiB` from vpr and `MB` from vivado are not the same unit.
Fixes #216 

### DIFF
--- a/symbiflow.py
+++ b/symbiflow.py
@@ -141,6 +141,8 @@ class VPR(Toolchain):
 
             self.run_steps()
             self.add_runtimes()
+            self.add_wirelength()
+            self.add_maximum_memory_use()
 
     def get_tool_params(self):
         if self.params_file:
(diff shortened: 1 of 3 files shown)


acomodi:
@kamilrakoczy There seem to be an issue with the memory usage extraction of Vivado

kamilrakoczy:
@acomodi I couldn't reproduce this issue locally (`make run-all` succeeded without any error). I think this is related to vivado version. I added more debug info and I will try to find what is the problem.

kamilrakoczy:
@acomodi I found and fixed the problem, now all tests are passing.

PR #219, opened 2020-08-13
Exhaust: Argument Fix and --board

ryancj14:
1. Gives the `--project` and `--toolchain` arguments choices like their counterparts in `fpagperf.py`. This way, `exhuast.py` can recognize bad arguments and print out the user's options.

2. Creates and adds functionality to the `--board` argument. This means you can now specify a board or boards you want to be tested by the `exhuast.py` run.


ryancj14:
This one is ready to merge if we are happy with the changes. I think the added board argument would be useful for both research and testing purposes.

acomodi:
LGTM

PR #223, opened 2020-08-17
Exhaust `--overwrite` argument

ryancj14:
Added an argument that overwrites previous builds to re-run any builds that previously existed. Where it would previously give False for 'overwrite' when calling the fpgaperf run function, now it provides self.overwrite, which is true if --overwrite is specified in the command and false otherwise.


kamilrakoczy (approved):
LGTM

ryancj14:
Ready for merge.

Issue #215, opened 2020-08-11
Transition to use an automated circuit/toolhain matrix generation


acomodi:
Currently, each project.json file specifies which toolchains are allowed and the various constraints files.

As https://github.com/SymbiFlow/fpga-tool-perf/issues/214 gets solved we should potentially transition to have the run tasks generated automatically without having to explicitly state what toolchains are available for a specific test.

Issue #214, opened 2020-08-11
Update all VPR tests to use XDC placement constraints


acomodi:
Having different constraints files for each toolchain for a specific circuit can be confusing and can lead to error.

PR #227, opened 2020-08-20
Exhaust Task List Generation

ryancj14:
These reflect changes requested by @acomodi and has to do with issue #215.

These are simply some initial changes and not ready to be merged. I would like some feedback and then discussion on how to modify or continue this work in relation to #214 and #215

exhaust.py is in fact functional (though it doesn't pass the kokoro tests) with the changes I have made so far. It seems like this will allow us to remove the toolchain specs within `project` files as soon as toolchains become functional with all projects and fpgaperf.py's code is updated. Also only tests the projects whose constr file names are simply "board.extension" (e.g. basys3.xdc, arty.pcf).

Now relies on the vendor specs in each project file, mostly to keep certain board and toolchain combos clear of one another.

### DIFF
--- a/fpgaperf.py
+++ b/fpgaperf.py
@@ -351,6 +351,12 @@ def env_ready():
     return True
 
 
+def verify_constraint(project, board, extension):
+    board_file = board + "." + extension
+    path = os.path.join(src_dir, project, 'constr', board_file)
+    return os.path.exists(path)
+
+
 def get_constraint(project, board, constr_list, extension):
     constr_file = [v for v in constr_list if v.endswith(extension)]
 
--- a/other/vendors.json
+++ b/other/vendors.json
@@ -1,7 +1,7 @@
 {
     "xilinx": {
         "boards": ["arty", "basys3", "zybo", "nexys"],
-        "toolchains": ["vivado", "vivado-yosys", "vpr", "vpr-fasm2bels", "nextpnr-xilinx", "nextpnr-xilinx-fasm2bels"]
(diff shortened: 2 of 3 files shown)


ryancj14:
Question: Some toolchains in the `project` .json files have multiple constraint files, but as far as I can tell, only one is specified as compatible according to tasks.py's MANDATORY_CONSTRAINTS list. What is the reasoning behind having multiple different constraint files for a single given project/toolchain/board combo and does it use them all?
Example in `baselitext.json`:
```
"vpr": {
    "arty": ["arty.pcf", "arty.sdc", "arty-vpr.xdc"]
},
"vpr-fasm2bels": {
    "arty": ["arty.pcf", "arty.sdc", "arty-vpr.xdc"]
}
```

acomodi:
@ryancj14 So, these set of constraints are there for three different reasons, and for the time being, all of them need to be separated, as there is no support yet to combine them all together in one set of constraints only.

- PCF: this defines the IO mapping, and it is used to bind a top level io-signal to an IOPAD location. The use of the PCF is being removed to have the IO constraints directly in the XDC in [this PR](https://github.com/SymbiFlow/fpga-tool-perf/pull/222).
- XDC: this is currently used to provide the IOSTANDARD for the IO pads. The IOSTANDARD defines how a specific IO block should be configured
- SDC: These are clock constraints used to drive the router and the placer to correctly perform timing driven P&R to reach the required clock frequencies. There is a WIP to have this automatically inferred during synthesis.

PR #231, opened 2020-08-27
Add quicklogic toolchain

kamilrakoczy:
This PR adds installation of `quicklogic toolchain` in separate conda env,
adds following boards:

-  `quickfeather-dev-board`
-  `chandalar` 

adds following projects:

- `jpeg_qnr`
- `irr`
- `rsdecoder`


Makefile
```diff
@@ -25,9 +25,17 @@ env:: | $(CONDA_ENV_PYTHON)
 	git submodule init
 	git submodule update --init --recursive
 	mkdir -p env/symbiflow
+	wget -q https://github.com/QuickLogic-Corp/quicklogic-fpga-toolchain/releases/download/v1.1.0/Symbiflow_v1.1.0.gz.run
+	mkdir env/quicklogic
```
mkurc-ant, line 29: Should be `mkdir -p ...`. This way `make env` doesn't fail when run second time.

Makefile
```diff
@@ -25,9 +25,17 @@ env:: | $(CONDA_ENV_PYTHON)
...
+	chmod +x Symbiflow_v1.1.0.gz.run
+	INSTALL_DIR=${PWD}/env/quicklogic ./Symbiflow_v1.1.0.gz.run
+	rm Symbiflow_v1.1.0.gz.run
+	source ./env/quicklogic/conda/bin/activate && pip3 install asciitable terminaltables colorclass "git+https://github.com/antmicro/edalize.git@quicklogic#egg=edalize"
```
mkurc-ant, line 33: There should be only one `&` otherwise all that pip stuff gets installed outside the env.

Makefile
kamilrakoczy: Fixed

Makefile
```diff
@@ -25,9 +25,17 @@ env:: | $(CONDA_ENV_PYTHON)
 	git submodule init
 	git submodule update --init --recursive
 	mkdir -p env/symbiflow
+	wget -q https://github.com/QuickLogic-Corp/quicklogic-fpga-toolchain/releases/download/v1.1.0/Symbiflow_v1.1.0.gz.run
+	mkdir -p env/quicklogic
+	chmod +x Symbiflow_v1.1.0.gz.run
+	INSTALL_DIR=${PWD}/env/quicklogic ./Symbiflow_v1.1.0.gz.run
+	rm Symbiflow_v1.1.0.gz.run
+	source ./env/quicklogic/conda/bin/activate && pip3 install asciitable terminaltables colorclass "git+https://github.com/antmicro/edalize.git@quicklogic#egg=edalize"
+	sed -i "s/\/home\/kkumar\/symbiflow-arch-defs\/build\/env\/conda\/bin\/python3/\/usr\/bin\/env python3/" ./env/quicklogic/install/bin/python/qlfasm #fix path to interpreter, see https://github.com/QuickLogic-Corp/quicklogic-fpga-toolchain/issues/16
```
acomodi, line 34: I am doubtful on this procedure. What happens is that the conda environment gets generated using a different methodology (instead of using the [make-env repository](https://github.com/SymbiFlow/make-env)).

I understand that quicklogic uses a different yosys and VTR versions, plus other python libraries not present in the original conda requirements.txt. I am unaware whether we can modify the conda configuration file to set two different conda environments, but, if this is doable, we should set the Quicklogic environment using the configuration files.

conf/requirements.txt
```diff
@@ -12,5 +12,5 @@ pandas
 yapf
 git+https://github.com/SymbiFlow/fasm.git#egg=fasm
 git+https://github.com/SymbiFlow/xc-fasm.git#egg=xc-fasm
-git+https://github.com/SymbiFlow/edalize.git@symbiflow#egg=edalize
+git+https://github.com/antmicro/edalize.git@quicklogic#egg=edalize
```
acomodi, line 15: The edalize changes to allow having the quicklogic toolchain should probably be merged in the symbiflow branch of the edalize fork.

acomodi (changes requested):
Comments in.
I am unsure on generating the conda environments in two different ways. This may lead to conflicts/inconsistencies. I would rather opt for using the same way of generating the conda environment using the make-env library.

Makefile
kamilrakoczy: Done, I changed installation to use make-env. Now when you are downloading environment, you need to specify which one: symbiflow or quicklogic:
```
TOOLCHAIN=symbiflow make env
TOOLCHAIN=quicklogic make env
```
By default (when toolchain is not specified explicit before make command) it is only downloading symbiflow.

Makefile
```diff
@@ -22,12 +27,14 @@ third_party/make-env/conda.mk:
...
-	mkdir -p env/symbiflow
+	mkdir -p env/${TOOLCHAIN}
 	wget -O ${SYMBIFLOW_ARCHIVE} ${SYMBIFLOW_URL}
-	tar -xf ${SYMBIFLOW_ARCHIVE} -C env/symbiflow
+	tar -xf ${SYMBIFLOW_ARCHIVE} -C env/${TOOLCHAIN}
 	rm ${SYMBIFLOW_ARCHIVE}
+ifeq ("${TOOLCHAIN}", "quicklogic")
+	cp -r env/conda/envs/quicklogic/share/yosys/* env/conda/envs/quicklogic/share/
+	sed -i "s/\/home\/kkumar\/symbiflow-arch-defs\/build\/env\/conda\/bin\/python3/\/usr\/bin\/env python3/" env/${TOOLCHAIN}/install/bin/python/qlfasm #fix path to interpreter, see https://github.com/QuickLogic-Corp/quicklogic-fpga-toolchain/issues/16
+endif
```
acomodi, line 37: This has two patches, one related to a fixed path (https://github.com/QuickLogic-Corp/quicklogic-fpga-toolchain/issues/16), and the other to a yosys issue for which the share directory cannot be found.

It would be better to have a FIXME or TODO before this block of code explaining why it is necessary

acomodi (changes requested):

Makefile
acomodi: Perfect, looks better now, thanks

conf/symbiflow/requirements.txt
```diff
@@ -12,5 +12,5 @@ pandas
 yapf
 git+https://github.com/SymbiFlow/fasm.git#egg=fasm
 git+https://github.com/SymbiFlow/xc-fasm.git#egg=xc-fasm
-git+https://github.com/SymbiFlow/edalize.git@symbiflow#egg=edalize
+git+https://github.com/antmicro/edalize.git@quicklogic#egg=edalize
```
acomodi, line 15: This change is not required anymore as https://github.com/SymbiFlow/edalize/pull/56 is now merged

kamilrakoczy: done

Makefile
kamilrakoczy: done

acomodi (approved):
LGTM

src/jpeg-qnr/div_su.v
```diff
@@ -0,0 +1,157 @@
+/////////////////////////////////////////////////////////////////////
```
mithro, line 1: This should be in third party

PR #234, opened 2020-09-14
Changes for visualization

kamilrakoczy:
This PR adds changes for visualization difference between different toolchains/boards/seeds.

1. It adds 2 more Makefile targets:

    - make run-all-multiple-samples-tests - this target runs all tests 2 times to make sure tools are deterministic

    - run-all-seedable-tests - this target runs all vpr tests with 20 randomly generated seeds between 1 and 2147483647 (max seed value for vpr)

2. Changes default frequency unit to MHz from Hz.

3. Fixes generation of ``fmax`` for vivado in various projects.

4. Adds ``run-all-multiple-samples-tests`` and ``run-all-seedable-tests`` to continuous build type.

5. Adds vendor and splits toolchain into ``synthesis_tool`` and ``pr_tool`` in ``meta.json``:


acomodi (approved):
LGTM, merging

PR #235, opened 2020-09-21
Exhaust Task List Generation

### LINKED ISSUE #215: Transition to use an automated circuit/toolhain matrix generation

kamilrakoczy:
Supersede of #227.
Fixes #215

This PR:

- Updates symbiflow and quicklogic toolchain versions.

- Adds ``symbiflow_a200t`` toolchain to ``make env``.

- Changes ``exhaust.py`` to by default build all possible combinations of board/toolchain/project.

- Adds ``--only_required`` parameter to ``exhaust.py`` to build only boards that are defined in ``project.json`` as required for CI to be green.

- Adds ``nexys-video`` board. 

- Limits ``exhaust.py`` exception message lenght to 300 characters (without this, log from CI weights ~200MB).


acomodi (approved):
LGTM. Let's wait for CI

acomodi:
I see in CI that the test combination matrix is generated also for the following tests:

- multiple_tests
- parameter_tests
- seed_tests

I think it might be good to add the --only-prefix parameter when invoking these tests as well.

src/ibex/constr/arty-yosys-vivado.xdc
```diff
@@ -0,0 +1,18 @@
+## Based on https://github.com/Digilent/digilent-xdc/blob/master/Arty-A7-100-Master.xdc
```
acomodi, line 1: Why is this constraint called arty-yosys-vivado instead of arty.xdc only? I would guess that the same  constraint file would not work on other toolchains. If not, I think this can be removed.

src/picosoc/constr/basys3-nextpnr-xilinx.xdc
```diff
@@ -0,0 +1,80 @@
+# Clock Signal
```
acomodi, line 1: Same thing as above. Is this needed for nextpnr-xilinx only and it does not work for other toolchains?

acomodi (approved):
LGTM apart from some indentation issues.

I see some constraints that differ among toolchains. I assume there are some incompatibilities among them, It would be good to add a comment in each of the "special" constraints to report what XDC command generates the issue for not having a unique constraint file among all the toolchains.

Let's wait for CI.

src/ibex/constr/arty-yosys-vivado.xdc
kamilrakoczy: Yes, I left option, to use toolchain-specific constraint file, if for some reason, same constraint file won't work on all toolchains, but in this case, it was some leftover. I removed this file and updated ``arty.xdc`` to work on all toolchains

src/picosoc/constr/basys3-nextpnr-xilinx.xdc
kamilrakoczy: Same as above.

PR #236, opened 2020-09-22
symbiflow: fix bram calculation

acomodi:

For symbiflow targets, RAMB36 were not taken into account in the resource usage extraction.

Moreover, each RAMB36 will count as double, as underneath, it uses two RAMB18


acomodi:
@mithro  FYI

mithro:
Should there be some type of testing here?

PR #239, opened 2020-10-01
exhaust: set error code only when required test failed

kamilrakoczy:
This PR set error code, only when required test failed. This allows to run all tests without ``--only-required`` flag and error code (and CI status) will be ``1`` only when test that are marked as required will fail.

This will reduce CI build time, by not building the same test 2 times.


acomodi (approved):
LGTM, merging. Travis CI seems to be still running, but the build completed successfully.

PR #241, opened 2020-10-06
Fix resource utilization iob

acomodi:
This PR fixes the resource utilizaiton extration regarding IOBs.

Also:
- get last 1000 characters of error message when test fails
- allow timing failures for nextpnr
- print out failed required tests


acomodi:
Kokoro went green, even though it still shows a pending state. Merging

PR #247, opened 2020-10-14
Add separate env for nextpnr toolchain

### LINKED ISSUE #245: Create separate environment for different toolchains
To get to a point where FPGA-tool-perf is completely automated, with results published regularly, the current environment setup is not suitable.

In fact, the environment yamls are currently separated only between quick-logic and symbiflow. It would be good to apply the same separation for the nextpnr tools as well, such that each different tool environment is well separated from the others.

This allows for easily updating the environment configuration files automatically.

E.g. We should likely have another directory with the nextpnr environment as done here: https://github.com/SymbiFlow/fpga-tool-perf/tree/master/conf

wtatarski:
Fixes #245 

This PR adds separate environment for nextpnr toolchain.

### DIFF
--- a/.github/kokoro/steps/hostsetup.sh
+++ b/.github/kokoro/steps/hostsetup.sh
@@ -81,5 +81,6 @@ echo "----------------------------------------"
 	TOOLCHAIN=symbiflow make env
 	TOOLCHAIN=symbiflow-a200t make env
 	TOOLCHAIN=quicklogic make env
+	TOOLCHAIN=nextpnr make env
 	echo "----------------------------------------"
 )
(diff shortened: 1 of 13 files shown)


acomodi (changes requested):
I am thinking that there might be some good improvement to be made to the environment scripts, now that we have different environments for each toolchain.

Right now, there seems to be confusion around the fact that we have the environments to run the specific toolchains with the conda-packages, and we have also the environment to have the fpga-tool-perf infrastructure running correctly (python packages mostly), and these environments are currently mingled together.

I think that we should modify how the environments are handled differently. e.g.:

- Have one common environment for fpga-tool-perf with the python packages to correctly generate the tests file through edalize, generate reports, and so on. This would include, for instance some of the python packages in the requirements such as:
  - terminaltables
  - asciitable
  - simplejson
  - tqdm
  - etc.

This could be done by just having a basic python virtualenv instead of conda.

Than there is another set of environments which are specific to the toolchains:
- symbiflow
- quicklogic
- nextpnr

These must be generated using conda, and I think that we might transition to have a different environment script for each different toolchain. Each script will be given to edalize so that it can be sourced and used for running the test: https://github.com/SymbiFlow/fpga-tool-perf/blob/77ef38f3fbc4f8ef9bc9fe598680e69edfd74571/symbiflow.py#L150-L152

mithro:
FYI - Requirements and environment files can include other requirements / environment files.

acomodi (approved):
LGTM

PR #248, opened 2020-10-15
ci: equal presubmit and continuous CIs script

acomodi:


mithro:
If CI passes, go ahead and merge!

PR #250, opened 2020-10-19
symbiflow: fix vpr frequencies

acomodi:

### DIFF
--- a/src/blinky/constr/arty.sdc
+++ b/src/blinky/constr/arty.sdc
@@ -1,1 +1,1 @@
-create_clock -period 10.0 \$abc\$2843\$iopadmap\$clk_i
+create_clock -period 10.0 clk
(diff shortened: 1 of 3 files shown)


src/blinky/constr/arty.sdc
```diff
@@ -1 +1 @@
-create_clock -period 10.0 \$abc\$2843\$iopadmap\$clk_i
+create_clock -period 10.0 clk
```
mithro, line 1: What does the XDC file say?

acomodi: The name of the input clock pin is `clk_i`. but the design contains an explicit BUFG which renames the system clock to `clk`.

Currently we do not have yet enabled the XDC parsing plugin yet in fpga-tool-perf. Once the plugin is enabled, we can drop the use of SDC.

PR #252, opened 2020-10-20
nexys: add missing clock constraints

acomodi:

### DIFF
--- a/project/iir.json
+++ b/project/iir.json
@@ -10,5 +10,5 @@
     "vendors": {
         "quicklogic": ["chandalar", "quickfeather"]
     },
-    "required_toolchains": ["quicklogic"]
+    "required_toolchains": []
 }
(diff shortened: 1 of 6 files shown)


mithro (approved):

project/iir.json
```diff
@@ -10,5 +10,5 @@
     "vendors": {
         "quicklogic": ["chandalar", "quickfeather"]
     },
-    "required_toolchains": ["quicklogic"]
```
mithro, line 13: This should be a seperate pull request?

acomodi: I have added a strict check to have no missing frequencies in the meta.json. This test is not able to produce frequency outputs, as there isn't a real clock net driving logic (this is a dummy test, not meant to run on HW as far as I have learned).

PR #254, opened 2020-10-22
Add arty-a7-100t

### LINKED ISSUE #246: Enable the arty a7 100T board
With the availability of the artix 100T device, we can enable tests on the arty a7 board as well.

This translates into:
- adding the arty_a7-100T board specification
- adding constraint files to the existing tests

wtatarski:
Fixes #246


src/baselitex/constr/arty-a35t-nextpnr-xilinx.xdc
```diff
@@ -0,0 +1,310 @@
+set_property LOC H5 [get_ports {led[0]}]
```
mithro, line 1: Why is the xdc specific to nextpnr?

acomodi: There seems to be an issue with reading XDC constraints related to the iobank (to get the INTERNAL_VREF). That line has been commented out for the nextpnr-xilinx test.

PR #255, opened 2020-10-27
Adding configuration for nightly run.

mithro:
Signed-off-by: Tim 'mithro' Ansell <me@mith.ro>


mithro:
@acomodi - The Google side has been set up, this is the side needed in fpga-tool-perf. Once landed and confirmed working we can modify the script run so it isn't identical to the continuous one.

mithro:
@litghost - FYI

acomodi (approved):
LGTM. As soon as the automatic fetch of symbiflow-arch-defs artifacts is in place we can proceed  by changing the script.

PR #257, opened 2020-10-28
Enable baselitex on nexys video

wtatarski:


HackerFoo (approved):

acomodi:
@mithro @HackerFoo I have generated a new design with VexRiscV_Linux.

This was achieved with https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1797 to land and the current symbiflow install packges uses the #1797 artifacts.

The design works at 100MHz, as a lower frequency does not work (even with Vivado), as the DDR memtest fails.

There is also an issue with the IDELAYCTRL, which I needed to temporary workaround.
Basically, the litex design contains now two banks of IDELAYs, one for ethernet and one for DDR. The problem is that there is only one instance of IDELAYCTRL which gets replicated by Vivado and not by Symbiflow, causing an issue during placement constraints generation. I will open an issue about this on symbiflow-arch-defs.

The bitstream is able to load Linux, but there seem to be problems with ethernet, as the loading process does not properly work and, once booted with `lxterm`, pinging the host at `192.168.1.100` produces no effect.

acomodi:
@HackerFoo @mithro I got the ethernet working, as a result of fixing IDDR and ODDR primitives in fasm2bels: https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/36

To get this merged and tested we need to merge the following PRs:
- [Yosys-Plugins SDC fix](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/57): This will require an update in symbiflow-arch-defs as well. This solves the issue for which we are getting inexistent clocks in the SDC
- [IDDR/ODDR fix](https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/36)

With these fixes Linux can boot from ethernet successfully.

acomodi:
@HackerFoo I have updated the test sources and applied some fixes to have the baselitex design working for 200T:

1. https://github.com/SymbiFlow/yosys-symbiflow-plugins/issues/64 This issue causes to have some missing clock constraints which result in VPR to not optimize the CP during P&R. To overcome this I have transitioned to use the manual SDC file. Once the issue is fixed and the package updated in archdefs, we can go back at having the XDC constraints only.

2. LOC constraints on the PLLs: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1750

3. Duplication of the IDELAYCTRL: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1804

With these workarounds, the design builds successfully, at least locally.

acomodi:
@HackerFoo There is an outstanding issue with the ODDR preventing ethernet to work with the current status of this PR. Fix is here in arch-defs https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1870#pullrequestreview-549335661.

With that fix in, ethernet works fine, and Linux boots

conf/symbiflow/environment.yml
```diff
@@ -1,15 +1,16 @@
...
-  - symbiflow-vtr=8.0.0.rc2_5415_gd6d69ff92=20201120_180018
-  - prjxray-tools
-  - prjxray-db
-  - python=3.8
+  - litex-hub::capnproto-java=0.1.5_0012_g44a8c1e=20201104_165332
+  - litex-hub::gcc-riscv64-elf-newlib=9.2.0=20201119_154229
+  - litex-hub::symbiflow-yosys-plugins=1.0.0_7_g59ff1e6_23_g3a95697_17_g00b887b_0194_g40efa51=20201120_145821
+  - litex-hub::prjxray-tools=0.1_2697_g0f939808=20201120_145821
+  - litex-hub::prjxray-db=0.0_0239_gd87c844=20201120_145821
+  - litex-hub::vtr=v8.0.0_3011_gb0223dc59=20201202_112618
```
HackerFoo, line 11: Is this supposed to be synchronized with symbiflow-arch-defs somehow? This is not the same as the version used by symbiflow-arch-defs (currently d912bdb88)

acomodi: Currently, in CI we are using the environment.yml coming from the symbiflow-arch-defs install package, and not this one. I am unsure how to deal with the update of this static environment.yml, for now it needs to be manual.

To override which environment variable to use (at least when building in conda) you can do:
```
ENVIRONMENT_FILE=<custom_env> make env
```

the default points at the one in `conf/symbiflow/environment.yml`.

PR #266, opened 2020-11-13
exhaust.py - saving table output to a text file

ryancj14:
I've found the table output to be interesting, and to be able to reference the output of the last exhaust.py run could be useful. This saves the table to a text file called `exhaust_output.txt` inside of whichever directory is set by `--out-prefix`, the default and usual location being `build/_exhaust-runs/exhaust_output.txt`.

### DIFF
--- a/exhaust.py
+++ b/exhaust.py
@@ -60,6 +60,7 @@ def print_summary_table(
             'Build N.', 'Options'
         ]
     ]
+    table2_data = table_data.copy()
     passed = failed = 0
     build_status = True
     build_count = 0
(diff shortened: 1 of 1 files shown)


ryancj14:
This pr is ready to merge. Let me know if it needs to be different or doesn't seem to be a needed change.

ryancj14:
I will no longer be working in the realm of student research related to SymbiFlow, so you can do with this PR as you wish. It is ready to be merged if you would like. Thanks.

mithro:
@acomodi - Any reason you didn't review this?

acomodi:
@mithro I think I have lost track of this PR, I'll take a look at it

acomodi (changes requested):
I believe that duplication of code to save the table data in an output file might not be necessary.

A possible solution might be to add an argument to determine whether the table will be displayed or saved on file and avoid duplicating code.

Given that this PR is outdated, I'll close it for now.

PR #280, opened 2020-11-27
nextpnr: Use Conda packages from LiteX-Hub channel

ajelinski:
The packages have been also updated to their latest versions.

### DIFF
--- a/conf/nextpnr/environment.yml
+++ b/conf/nextpnr/environment.yml
@@ -11,6 +11,9 @@ dependencies:
   - litex-hub::prjxray-db=v0.0_239_gd87c844=20210108_144229
   - litex-hub::prjxray-tools=0.1_2697_g0f939808=20201120_145821
   - libboost=1.73.0
+  # swig is required to build the rapidyaml package
+  # It can be removed once that is available in PyPi
+  - swig
   - pip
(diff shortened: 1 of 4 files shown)


ajelinski:
The PR title and body have been updated because `conf/symbiflow/environment.yml` already uses packages from the LiteX-Hub channel on the `master` branch. Therefore the commit has been modified to only change `conf/nextpnr/environment.yml` and placed on top of the current `master`.

ajelinski:
I have rebased branch onto the current master. Packages' versions were modified to have the packages as close to the ones used in the latest successfully tested commit on `master` as possible. I hope this will make the kokoro testing successful.

ajelinski:
I've added a commit to lock `python-fpga-interchange` package revision to the one before using `rapidyaml` which turns out to be very problematic: https://github.com/SymbiFlow/python-fpga-interchange/issues/19 .

This should also fix CI on master as it fails exactly because of it.

ajelinski:
I guess it should @mithro. I can see there even is a PR pending: https://github.com/SymbiFlow/prjxray/pull/1550 but it seems forgotten. If it isn't merged until I get this PR green, I'll add the URL as a comment.

mithro:
Do you need me for something here?

PR #282, opened 2020-12-02
Adding JSON and Verilog Formatters

ryancj14:
I think it is important for the repository to be neat and understandable, so I made some quick formatting corrections to verilog and json files. This should probably be a merge-when-green.


acomodi:
@ryancj14 Thanks for this PR, much appreciated. There is an issue with the pip upgrade that prevents the environment to be correctly generated. There is a fix in this PR: https://github.com/SymbiFlow/fpga-tool-perf/pull/285. Can you please rebase on top of that PR, so that CI can run correctly?

ryancj14:
This one should be good to merge.

litghost:
Would you mind including how you did the reformatting in the PR?  E.g. extend https://github.com/SymbiFlow/fpga-tool-perf/blob/master/Makefile#L58 ?

litghost (changes requested):
We do apprecate this change, but formatting needs to be automated. Would it be possible to add how you formatted these files to the root Makefile?

Thanks.

ryancj14:
I manually formatted most of these. I just figured that a few changes would make it that much better. It makes sense that they should be automated though. I believe yapf only works for python files. If there is another method of automated formatting for json and verilog files? I would be happy to implement some of that if someone knows of a quality way to do it.

litghost:
> I manually formatted most of these. I just figured that a few changes would make it that much better. It makes sense that they should be automated though. I believe yapf only works for python files. If there is another method of automated formatting for json and verilog files? I would be happy to implement some of that if someone knows of a quality way to do it.

json can be formatted with `python -mjson.tool`.  I believe @hzeller can answer about Verilog formatting.

hzeller:
Yes, Verible provides a (System)Verilog formatter: https://github.com/google/verible/tree/master/verilog/tools/formatter

ryancj14:
The mjson and verible formatters worked on my machine, but aren't passing the CI tests, because I've added to the make format command in the makefile. I believe the CI tests need to be updated to include formatting using mjson and verible. Is that something I can do on this pull request in ci.yml, or does it need to be done by someone else?

Makefile
```diff
@@ -54,10 +54,18 @@ install_quicklogic:
...
+VERILOG_SRCS=$(shell find . -name "*.v" -not -path "./third_party/*" -not -path "./env/*" -not -path "./conf/*")
+
+format:
+	yapf -i ${PYTHON_SRCS}
+	for i in ${JSON_SRCS}; do \
+		echo "$$(mjson $$i)" > $$i; \
+	done
+	for i in ${VERILOG_SRCS}; do \
+		echo "$$(verible-verilog-format $$i)" > $$i; \
+	done
```
acomodi, line 68: Given that mjson and verible are added to the conda environment, I believe that the environment needs to be activated before formatting (I see in CI that verible and mjson cannot be found)

ryancj14: How do I go about activating the environment in the CI test?

acomodi: I think sourcing the env.sh in the ci.yml before calling `make format` should work

ryancj14: Ok, that did work. I added make env and then source env.sh.

ryancj14:
I'm not sure why the build is failing, but it looks like all the other pull requests not by a bot are failing too. Is there anything I need to fix here to help it succeed?

ryancj14:
This pull request will become irrelevant with the integration of symbiflow-common-config, so I'll close it now.

PR #283, opened 2020-12-02
Fixed --list-combinations Functionality

ryancj14:
With recent changes, the `--list-combinations` functionality broke. I got it up and running again. It now uses the `"vendor"` and `"required_toolchains"` sections of each project's json file to determine whether a project/toolchain/board combination is "Supported" or "Missing".


mithro:
Should we have a test for this so it doesn't break again in the future?

acomodi:
Yeah, I think this check could be added to the GH actions CI, as well as a check for https://github.com/SymbiFlow/fpga-tool-perf/pull/284/files. For the latter we would need to generate the conda environment for all the toolchains as well.

PR #284, opened 2020-12-02
Fixed --check-env Functionality

ryancj14:
Something got moved preventing --check-env working for certain toolchains. This was easily fixed by importing have_exec to those toolchain's python files.


acomodi:
@ryancj14 Merged, thanks for the fix.

PR #294, opened 2020-12-09
exhaust: allow selecting how many cpus to use

acomodi:

### DIFF
--- a/.github/kokoro/continuous.cfg
+++ b/.github/kokoro/continuous.cfg
@@ -29,3 +29,8 @@ env_vars {
   key: "KOKORO_DIR"
   value: "fpga-tool-perf"
 }
+
+env_vars {
+  key: "NUM_CORES"
+  value: "10"
+}
--- a/.github/kokoro/nightly.cfg
+++ b/.github/kokoro/nightly.cfg
@@ -29,3 +29,8 @@ env_vars {
   key: "KOKORO_DIR"
   value: "fpga-tool-perf"
 }
+
+env_vars {
+  key: "NUM_CORES"
+  value: "10"
+}
--- a/.github/kokoro/nightly.sh
+++ b/.github/kokoro/nightly.sh
@@ -8,12 +8,18 @@ source ./.github/kokoro/steps/hostsetup.sh
 source ./.github/kokoro/steps/hostinfo.sh
 source ./.github/kokoro/steps/git.sh
 
+if [ -z $NUM_CORES  ]; then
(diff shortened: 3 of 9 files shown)


mithro:
@acomodi - It might be good that when running performance tests the tool lock everything to a given CPU core? https://unix.stackexchange.com/questions/23106/how-to-limit-a-process-to-one-cpu-core-in-linux

mithro:
https://man7.org/linux/man-pages/man1/taskset.1.html

acomodi:
@mithro All right, I verified this works on my local machine. But, given the amount of tests we are running, locking everything to one CPU might let the overall run-time of CI to spike. For instance, it took ~3x time on my local machine to run all the blinky test using one locked CPU instead of multiple CPUs.

I think we could check how the reduction in NUM_CORES added with this PR works out, and then, if we still see high noise w.r.t. run-time in the final graph, we could think of locking onto a single CPU

mithro:
I mean lock each task that is run (rather than the whole FPGA tool perf).

acomodi:
I am unsure how to do that with the multiprocess python library. What is happening in toolperf is that the task list is passed to the multiprocessing Pool and python dispatches all the tasks. Then the OS is handling all the job allocations to the various cores.

mithro:
http://sorami-chi.hateblo.jp/entry/2016/04/29/000000 apparently?

mithro:
And https://bugs.python.org/msg281931 maybe?

mithro:
I wouldn't necessarily block this pull request on the locking to a single CPU stuff.

acomodi:
Hmm, I bumped into the first link, but I do not think it might be helpful here, as it requires modifying the `multiprocessing` library. Even though a feature that enables the user to lock the process within a specific CPU core would be a nice improvement to the library itself.

Regarding the psutil approach, even though we can alter the CPU affinity of a process, in our case the Pool is handling all of that.

I am thinking that we might overcome this problem by allowing the worker function to call a subprocess with taskset and the call to fpgaperf.py with all the required parameters of the task?
I am unsure though if the Pool does some sort of re-organization of the processes under the hood. I'll need to investigate some more. Maybe there are other multiprocessor/multithreading libraries that could potentially help us here.

mithro:
Running a subprocess command inside a taskset might be the right solution?

acomodi:
@mithro I have opened an issue about this: https://github.com/SymbiFlow/fpga-tool-perf/issues/296. I will open a follow-up PR as soon as I have something ready. I'd merge this PR for the time being

PR #299, opened 2020-12-18
[DNM] Switch vtr to vtr-optimized.

HackerFoo:
Use vtr-optimized now that it's in symbiflow-arch-defs.


HackerFoo:
@acomodi Does this file still need to be updated? If not, maybe it should be removed.

acomodi:
@HackerFoo we can actually remove the environment.yml from `conf/symbiflow` and set to grab the default one from the install package

HackerFoo:
If I understand https://github.com/SymbiFlow/fpga-tool-perf/pull/290, this file is only overridden by the CI scripts.

I think the default should be to use the environment.yaml from the install package. Otherwise it would be surprising to try fpga-tool-perf locally and have it break because all the packages are out of date.

acomodi:
Indeed, we need to set to use the environment.yaml from the install package by default even for non-CI systems.

HackerFoo:
Alright, I'm going to mark this PR do not merge, because I want to see the results of CI, and then close it afterwards, because there is no need to make this update.

PR #300, opened 2020-12-21
Add dependencies for FASM ANTLR parser.

HackerFoo:
These Ubuntu packages are required to build the faster ANTLR FASM parser. Without them, the `fasm` package will fall back to the previous parser implementation.

- default-jre-headless
- uuid-dev

### DIFF
--- a/.github/kokoro/steps/hostsetup.sh
+++ b/.github/kokoro/steps/hostsetup.sh
@@ -63,6 +63,8 @@ sudo apt-get install -y \
         python3-virtualenv \
         python3-yaml \
         virtualenv \
+        default-jre-headless \
+        uuid-dev \
 
 if [ -z "${BUILD_TOOL}" ]; then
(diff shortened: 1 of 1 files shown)


acomodi (approved):
LGTM.

AFAIK, the ANTLR binaries would need some extra inputs to get correctly generated as discussed in https://github.com/SymbiFlow/symbiflow-examples/issues/108.

Does the same happen here or these changes are enough to get everything compiled correctly?

HackerFoo:
@acomodi These changes should be enough; they are the same as what is used in CI for `fasm`, except `libantlr4-runtime-dev` is not required for the static configuration.

HackerFoo:
Superseded by https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1908

I've moved these dependencies to the `environment.yml` inherited from `symbiflow-arch-defs`.

PR #313, opened 2021-01-07
Add VivadoYosysUhdm toolchain

kamilrakoczy:
This PR adds new toolchain using vivado + yosys with uhdm frontend.

### DIFF
--- a/conf/symbiflow/environment.yml
+++ b/conf/symbiflow/environment.yml
@@ -3,6 +3,7 @@ channels:
   - defaults
   - litex-hub
 dependencies:
+  - litex-hub::yosys-uhdm=0.5_8401_g30b87e190
   - python=3.7
   - pip
   - pip:
--- a/toolchains/vivado.py
+++ b/toolchains/vivado.py
@@ -108,6 +108,15 @@ class Vivado(Toolchain):
 
                 vivado_settings = os.getenv('VIVADO_SETTINGS')
 
+                vivado_options = {
+                    'part': chip,
+                    'synth': self.synthtool,
+                    'vivado-settings': vivado_settings,
+                    'yosys_synth_options': self.synthoptions,
+                }
+                if hasattr(self, 'library_files'):
+                    vivado_options['library_files'] = self.library_files
+
                 self.edam = {
                     'files': self.files,
                     'name': self.project_name,
@@ -121,16 +130,9 @@ class Vivado(Toolchain):
                                     'default': 1,
                                 },
                         },
-                    'tool_options':
-                        {
-                            'vivado':
-                                {
-                                    'part': chip,
-                                    'synth': self.synthtool,
-                                    'vivado-settings': vivado_settings,
-                                    'yosys_synth_options': self.synthoptions,
-                                }
-                        }
+                    'tool_options': {
+                        'vivado': vivado_options
+                    }
                 }
                 self.backend = edalize.get_edatool('vivado')(
                     edam=self.edam, work_root=self.out_dir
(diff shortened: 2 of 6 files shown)


toolchains/vivado.py
```diff
@@ -329,3 +331,15 @@ def add_runtimes(self):
...
+class VivadoYosysUhdm(VivadoYosys):
+    '''Vivado PnR + Yosys synthesis using uhdm frontend'''
+    def __init__(self, rootdir):
+        VivadoYosys.__init__(self, rootdir)
+        self.synthtool = 'yosys'
+        self.synthoptions = [
+            '-iopad', '-arch xc7', '-flatten', 'frontend=surelog'
+        ]
+        self.toolchain = 'yosys-vivado-uhdm'
+        self.library_files = f"{self.rootdir}/env/conda/envs/symbiflow-env/share/uhdm-yosys/xilinx/cells_sim.v,{self.rootdir}/env/conda/envs/symbiflow-env/share/uhdm-yosys/xilinx/cells_xtra_surelog.v"
```
acomodi, line 345: Is there a way to not have to write this explicitly? Mainly because this requires accessing a specific conda environment. Maybe you could get the share directory with `yosys-config --datdir`

conf/symbiflow/environment.yml
```diff
@@ -3,6 +3,7 @@ channels:
   - defaults
   - litex-hub
 dependencies:
+  - litex-hub::yosys-uhdm=0.5_8401_g30b87e190
```
acomodi, line 6: Rethinking about this, I am worried that this yosys-uhdm replaces the yosys from the symbiflow package that is used for the VPR tests. 

I am not sure whether it would be better to actually have this in yet another separate environment

acomodi (changes requested):
Comments in

conf/symbiflow/environment.yml
kamilrakoczy: ``yosys-uhdm`` package uses ``PROGRAM_PREFIX`` and renames binary to ``uhdm-yosys``: https://github.com/hdl/conda-eda/blob/master/syn/yosys-uhdm/build.sh#L16

It should not replace yosys for other tests, but if you think it would be better to have separate environment for it, I can do it.

acomodi: All right, so, I believe that we can leave it as is, if the whole uhdm-yosys package lives separately from the original yosys one, I think we can leave it as is.

toolchains/vivado.py
kamilrakoczy: I cant use ``yosys-config`` as it returns path to datdir where this package was built, e.g.: ``/home/runner/work/conda-eda...``.

I've changed this, to only assume, that ``uhdm-yosys`` is installed in symbiflow-env, but if uhdm-yosys is in PATH, we are pointing to the library files from version in the PATH.

toolchains/vivado.py
```diff
@@ -329,3 +331,27 @@ def add_runtimes(self):
...
+            '-iopad', '-arch xc7', '-flatten', 'frontend=surelog'
+        ]
+        self.toolchain = 'yosys-vivado-uhdm'
+        # by default assume, uhdm-yosys is installed in symbiflow env
+        uhdm_yosys_path = f"{self.rootdir}/env/conda/envs/symbiflow-env/share"
+        for path in os.environ["PATH"].split(os.pathsep):
+            exe_file = os.path.join(path, "uhdm-yosys")
+            if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):
+                uhdm_yosys_path = path
+                break
```
acomodi, line 351: we may use `which uhdm-yosys` to get the location of the binary here, without assuming it from conda or the PATH

toolchains/vivado.py
```diff
@@ -329,3 +331,25 @@ def add_runtimes(self):
...
+        VivadoYosys.__init__(self, rootdir)
+        self.synthtool = 'yosys'
+        self.synthoptions = [
+            '-iopad', '-arch xc7', '-flatten', 'frontend=surelog'
+        ]
+        self.toolchain = 'yosys-vivado-uhdm'
+        uhdm_yosys_path = shutil.which("uhdm-yosys")
+        # by default assume, uhdm-yosys is installed in symbiflow env
+        if uhdm_yosys_path is None:
+            uhdm_yosys_path = f"{self.rootdir}/env/conda/envs/symbiflow-env/bin/uhdm-yosys"
```
acomodi, line 348: I think we should error out if the path is not there, as uhdm_yosys cannot be found, unless there is a strong reason to have this default assumption.

acomodi (approved):
LGTM, let's wait for CI

PR #317, opened 2021-03-09
make env: update base env with symbiflow environment

kamilrakoczy:
This PR tries to unify created environment between CI and user.
Up to now, kokoro CI was explicitly selecting environment file that is downloaded as part of symbiflow-arch-def (and it was not using ``conf/symbiflow/environment.yml`` at all), but when user follows README there is no information about selecting ``ENVIRONMENT_FILE`` explicitly when creating symbiflow environment, so for user, ``conf/symbiflow/environment.yml`` is used.

This PR changes target ``install_symbiflow`` to first install environment from ``conf/symbiflow/environment.yml`` and then updates it with ``env/symbiflow/environment.yml`` file. This way, we are installing dependencies from both files.

### DIFF
--- a/.github/kokoro/steps/hostsetup.sh
+++ b/.github/kokoro/steps/hostsetup.sh
@@ -85,7 +85,7 @@ echo "----------------------------------------"
 	echo
 	echo " Configuring conda environment"
 	echo "----------------------------------------"
-	ENVIRONMENT_FILE=env/symbiflow/environment.yml TOOLCHAIN=symbiflow make env
+	TOOLCHAIN=symbiflow make env
 	TOOLCHAIN=quicklogic make env
 	TOOLCHAIN=nextpnr make env
 	echo "----------------------------------------"
--- a/Makefile
+++ b/Makefile
@@ -36,12 +36,13 @@ include third_party/make-env/conda.mk
 
 env:: | $(CONDA_ENV_PYTHON)
 
-install_symbiflow:
+install_symbiflow: | $(CONDA_ENV_PYTHON)
 	mkdir -p env/symbiflow
 	curl -s ${SYMBIFLOW_LATEST_URL} | xargs wget -qO- | tar -xJC env/symbiflow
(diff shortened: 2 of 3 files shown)


kamilrakoczy:
Quicklogic tests are failing due to missing ``libtbb.so.2`` and it should be fixed after: https://github.com/hdl/conda-eda/pull/90

ajelinski:
Unfortunately https://github.com/hdl/conda-eda/pull/90 won't fix that issue in already built VTR packages. It will only prevent such problems in packages built after merging the PR. For more details, see https://github.com/hdl/conda-eda/issues/84#issuecomment-794264558 .

Only renaming `tbb` package for new and incompatible TBBv2021 is able to fix the root of the problem and I've added an issue to make such a change: https://github.com/conda-forge/tbb-feedstock/issues/81

You can simply mitigate it like in https://github.com/SymbiFlow/symbiflow-examples/pull/129/files#diff-0eee40a2465b95b175ff4339c6e8a2b4319282376eb4ac997aac8a6c5a32a86cR9 but don't use that comment.
If you want to state when it can be removed then the only options are:
* VTR package used was built after merging https://github.com/hdl/conda-eda/pull/90
* there's no `tbb-2021.1.1` package on Conda (e.g. it got reuploaded under a different name)

kamilrakoczy:
@acomodi I rebased PR to include pinned tbb version change.

Now all tests are passing, could you review this PR and check is this approach to make CI and user use same environment files is correct?

acomodi:
@kamilrakoczy I think the approach is ok. The only thing would be probably to have the "static" environment file containing only the additional package that you require (for instance the yosys-uhdm package). Than, all the other packages will be installed from the tarball. This way we make sure that the packages we are using (such as VTR, etc.) come from the installed toolchain environment file.

kamilrakoczy:
@acomodi I've removed all packages from static environment file that are also installed from tarball, but left python/pip, to make sure conda will install correct python version. Otherwise conda will install newest python version and it will conflict with packages installed from tarball.

PR #323, opened 2021-04-08
Add nextpnr fpga interchange

kowalewskijan:
This is the initial PR for `nextpnr-fpga_interchange` support in fpga-tool-perf.

Dependencies:
- [x] https://github.com/SymbiFlow/edalize/pull/94
- [x] https://github.com/hdl/conda-eda/pull/102

Known issues:
- [x] using `LOC` is not supported - we expect to have `PACKAGE_PIN` instead in XDC (https://github.com/YosysHQ/nextpnr/pull/683)
- [x] `create_clock` command is not supported (we'll have a separate `blinky_interchange` project without this command)

Statuses:
- [x] `.phys` files generation
- [x] `.fasm` files generation
- [x] `.bit` files generation
- [x] performance report for `fpga-interchange` variant


toolchains/symbiflow.py
```diff
@@ -1001,6 +1003,451 @@ def check_env():
...
+                        l_split = l.split()
+                        wirelen = l_split[l_split.index("wirelen") + 2]
+                        wirelen = ''.join(c for c in wirelen if c.isdigit())
+                        wirelen = int(wirelen)
+            return wirelen
+
+        route_log = os.path.join(self.out_dir, 'nextpnr.log')
+        self.wirelength = get_wirelength(route_log)
+
+    def max_freq(self):
```
acomodi, line 1226: I might be wrong, but different nextpnr arch binaries should have the same frequency logs, therefore it should be possible to get this by inheritance.

toolchains/symbiflow.py
```diff
@@ -1001,6 +1003,451 @@ def check_env():
         }
 
 
+class NextpnrFPGAInterchange(Toolchain):
```
acomodi, line 1006: We should have a generic Nextpnr class for all different architectures available, such that we do not duplicate code across archs.

toolchains/symbiflow.py
```diff
@@ -1001,6 +1003,451 @@ def check_env():
...
+        ret = {
+            "LUT": lut,
+            "DFF": dff,
+            "BRAM": bram,
+            "CARRY": carry,
+            "GLB": None,
+            "PLL": pll,
+            "IOB": iob,
+        }
+        return ret
```
acomodi, line 1337: We need to think on how to deal with resources in a more generic way, as the FPGA interchange is meant to work for any kind of architecture and FPGA family and not be bound to one of them only (e.g. xc7).

This implementation of the resources function assumes that we are dealing with and xc7 device, which may not be true as soon as other architectures are available.

toolchains/symbiflow.py
```diff
@@ -1001,6 +1003,451 @@ def check_env():
...
+        yosys_log = os.path.join(self.out_dir, 'yosys.log')
+        nextpnr_log = os.path.join(self.out_dir, 'nextpnr.log')
+
+        synth_times = self.get_yosys_runtimes(yosys_log)
+        impl_times = self.get_nextpnr_runtimes(nextpnr_log)
+
+        for t in synth_times:
+            self.add_runtime(t, synth_times[t])
+        for t in impl_times:
+            self.add_runtime(t, impl_times[t])
```
acomodi, line 1403: Add these to the parent Nextpnr class as these contain code that may be duplicated across `Nextpnr<Arch>` classes

acomodi (changes requested):
Comments in

toolchains/symbiflow.py
kowalewskijan: PTAL: [NextpnrGeneric](https://github.com/SymbiFlow/fpga-tool-perf/pull/323/files#diff-69ab290f878020ed67b1008338e82df43674e16b1e45733d5f0d8e04b6601141R585-R604)

toolchains/symbiflow.py
kowalewskijan: I think we can introduce the [resources_map](https://github.com/SymbiFlow/fpga-tool-perf/pull/323/files#diff-69ab290f878020ed67b1008338e82df43674e16b1e45733d5f0d8e04b6601141R585-R604) in the generic part where we could add arch and vendor dependent keywords related to resources.

toolchains/symbiflow.py
kowalewskijan: Moved to [NextpnrGeneric](https://github.com/SymbiFlow/fpga-tool-perf/pull/323/files#diff-69ab290f878020ed67b1008338e82df43674e16b1e45733d5f0d8e04b6601141R585-R604)

toolchains/symbiflow.py
kowalewskijan: This will be common for both Xilinx and FPGAInterchange variants, but for now we non-generic [overridden method](https://github.com/SymbiFlow/fpga-tool-perf/pull/323/files#diff-69ab290f878020ed67b1008338e82df43674e16b1e45733d5f0d8e04b6601141R1081-R1108)

kowalewskijan:
I've done some changes regarding migrating from `environment_script` to `EDALIZE_LAUNCHER` and using Yosys templates for customizable synthesis. Yosys templates are not in the master branch yet hence I've changed temporarily the edalize branch to `antmicro/nextpnr+fasm2bels` which has commits with yosys templates and enabled fasm2bels feature. Fasm2bels option will be submitted as a separate PR to edalize when the nextpnr PR will be merged. Currently the CI should fail for UHDM target as this part hasn't been upstreamed yet.

toolchains/symbiflow.py
```diff
@@ -971,36 +1008,180 @@ def yosys_ver():
...
+    # Fill the clock data with what we have...
+    def max_freq(self):
+        """Returns the max frequencies of the implemented design."""
+        log_file = os.path.join(self.out_dir, 'nextpnr.log')
+
+        clocks = dict()
+
+        with open(log_file, "r") as file:
+            for line in file:
+                if "target frequency" in line:
```
kowalewskijan, line 1115: We should probably leave the clock info as N/A

project/blinky_interchange.json
```diff
@@ -0,0 +1,14 @@
+{
```
acomodi, line 1: What is the difference between this and the blinky.json project? I think that, if there is only the addition of the required_toolchain, this project is not actually needed, and the blinky one can be used instead

src/yosys_templates/nextpnr-fpga_interchange_yosys_template.tcl
```diff
@@ -0,0 +1,20 @@
+yosys -import
```
acomodi, line 1: Why are these templates required?

toolchains/symbiflow.py
```diff
@@ -580,177 +583,223 @@ def check_env():
...
-                os.makedirs(self.out_dir, exist_ok=True)
+    def get_share_data(self):
+        out = subprocess.run(
+            ["find", ".", "-name", self.toolchain_bin], stdout=subprocess.PIPE
+        )
+        nextpnr_locations = out.stdout.decode('utf-8').split('\n')
+        nextpnr_location = None
+
+        for location in nextpnr_locations:
+            if "env/bin/{}".format(self.toolchain_bin) in location:
```
acomodi, line 648: the `env/bin` assumes that there is a special environment (e.g. conda in this case) that nextpnr is installed into. Might be worth removing the `env`?

acomodi (commented):
Generally LGTM, some comments in.

I think that we'd probably need to wait to have the edalize PR merged, and CI to go green.
I see that yosys-vivado started to fail, probably to some changes required with the different edalize version?

acomodi:
Superseded by https://github.com/SymbiFlow/fpga-tool-perf/pull/331. Closing

PR #343, opened 2021-08-19
Adjust results file

acomodi:
This PR does the following:

- fixes the regex for kokoro to upload the results files
- removes uninteresting data from the results and uses gzip to compress the data


mithro:
FYI - @QuantamHD

PR #357, opened 2021-10-02
Delete expired Let's Encrypt certificate.

mithro:
The old `/usr/share/ca-certificates/mozilla/DST_Root_CA_X3.crt` Let's Encrypt certificate that has expired.

The "best" solution seems to be to delete the certificate. This is fine because the new non-expired
`/usr/share/ca-certificates/mozilla/ISRG_Root_X1.crt` certificate is still available.

See the post at https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/ for more explanation.

### DIFF
--- a/.github/kokoro/steps/hostsetup.sh
+++ b/.github/kokoro/steps/hostsetup.sh
@@ -11,6 +11,29 @@ echo "----------------------------------------"
 
 echo
 echo "========================================"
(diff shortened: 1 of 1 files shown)


acomodi (approved):
LGTM, re-run kokoro as there seemed to be a connection error (not related to these changes)

PR #361, opened 2021-10-12
ci: move feature testing in GH actions rather than kokoro

acomodi:


acomodi:
> Yay!

FYI, for now there are only small tests performed in GH actions (e.g. custom parameters, seed, multiple runs). At the moment kokoro is still needed to run all the various tests

PR #366, opened 2021-10-29
Add baselitex nexus design for Lattice crosslink parts

tcal-x:
lifcl-40 works.   lifcl-17 works with current master nextpnr-nexus
but not with the version provided by Conda in this repo, so it's disabled.

The LiteX source was hand-modified to work with both lifcl-40 and -17,
by removing the PLL and driving sys_clock directly from the OSCA.

Signed-off-by: Tim Callahan <tcal@google.com>


mithro:
@acomodi - PTAL.

project/baselitex-nexus.yaml
```diff
@@ -0,0 +1,20 @@
...
+clocks:
+  PLL_0_P: 13.33333
+clock_aliases:
+  sys_clk:
+    - sys_clk
+    - sys_clk$glb_clk
+vendors:
+  lattice-nexus:
+    - lifcl-40
+#   - lifcl-17
```
acomodi, line 18: Maybe worth adding a comment on why this is not being run

acomodi (approved):
LGTM, with one small comment

project/baselitex-nexus.yaml
tcal-x: I made a change so that `nextpnr-nexus` uses its `router1` algorithm, which is what HPS uses, and now `oneblink` on `lifcl-17` also passes, so I've re-enabled it.

mithro: It seems like we should have runs for both router algorithms?

tcal-x: > It seems like we should have runs for both router algorithms?

Agreed, I have code for that.

*Currently* with router2, the design hits the infinite loop with design lifcl-17 and seed=1.   It's fine if the run times out and gets marked down as a failure, but I don't see any timeout / limit for each individual run.  @acomodi , will the `exhaust.py` script do the right thing and kill the one task, or will the never-ending job wreck the entire tool-perf run?

acomodi: At the moment the job will be never-ending if enters in a loop. We might add a feature to kill a job once a time limit is reached, to avoid infinite loops

acomodi (approved):
LGTM

PR #371, opened 2021-11-09
tasks: fix inconsistency in project name and project JSON file reference

acomodi:

### DIFF
--- a/other/boards.yaml
+++ b/other/boards.yaml
@@ -41,8 +41,8 @@ chandalar:
 lifcl-40:
   family: nexus
   device: LIFCL-40-9BG400CES
-  package: None
+  package: null
 lifcl-17:
   family: nexus
   device: LIFCL-17-8UWG72C
-  package: None
+  package: null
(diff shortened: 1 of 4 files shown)


other/boards.yaml
```diff
@@ -41,8 +41,8 @@ chandalar:
 lifcl-40:
   family: nexus
   device: LIFCL-40-9BG400CES
```
tcal-x, line 43: @acomodi , do you think `LIFCL-40-9BG400CES` should be broken into device and package?   The "9" is speed grade, and "ES" is engineering sample, so the last field is not strictly about the package, but for e.g. Xilinx parts the "package" is not strictly about the package either.

I'm fine either way.

mithro: I don't know how this matches to the usage in `fpga-tool-perf` but I would say that package is about how the silicon is broken out to real pins. You can have multiple different devices in the same package and multiple different packages for the same device.

acomodi: Yeah, this pretty much is the correspondence in `fpga-tool-perf`.

- family -> architecture type in terms of structure (e.g. UltraScale+, Series7, nexus, etc.)
- device -> number/disposition of LUTs, BRAMs, DSPs, etc. (a50t, a100t, etc.)
- package -> usable IO pins

A combination of the three is a `part`,

PR #380, opened 2021-11-22
[DNM] Runners tests

kgugala:

PR #378, opened 2021-11-22
gh-actions: add custom runners to parallelize jobs

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR is to enable a matrix computation of all the possible tests run by CI. It spawns one runner for each combination among all the possible valid tests combinations.

A follow-up PR would be the one to upload the results to a GCP or another location, and download from there the results to display in the dashboard.

PR #382, opened 2021-11-25
GH actions: add custom runners to parallelize jobs

acomodi:
This PR is built after the development performed in https://github.com/SymbiFlow/fpga-tool-perf/pull/380, https://github.com/SymbiFlow/fpga-tool-perf/pull/379 and https://github.com/SymbiFlow/fpga-tool-perf/pull/378.

It adds the possibility to run all the test suite in parallel, one task per runner, drastically reducing the CI cycle time. At the end of all the runs, the results are collected and aggregated and pushed to a GCP bucket.

The GH pages CI grabs all the results from the bucket and generates the dashboard, as it previously happened.


kgugala:
we should drop kokoro in this PR

acomodi:
> we should drop kokoro in this PR

Yeah, we may do this, and it should also happen on the kokoro side, as CI will still run and it won't find the kokoro configuration file to run the job.

acomodi:
@mithro As we transition to GH custom runners, the `countinuous`, `nightly` and `presubmit` instances of kokoro need to be removed to have green CI

mithro (commented):
Please land this in multiple steps;
 (a) Add the Github Actions stuff.
 (b) Wait a bit and confirm both are working together.
 (c) Get a Googler to remove the Kokoro config.
 (d) Remove the kokoro configurations.

acomodi:
@mithro Ok, we have restored Kokoro, we'll merge this one and check that everything is stable and proceed with point c) and d)

PR #383, opened 2021-12-23
Enhance Radiant support

acomodi:
This PR enhances the support for the radiant toolchain to use Edalize


src/baselitex-nexus/baselitex_nexus.v
```diff
@@ -4,7 +4,21 @@
...
+	output reg user_led4,
+	output reg user_led5,
+	output reg user_led6,
+	output reg user_led7,
+	output reg user_led8,
+	output reg user_led9,
+	output reg user_led10,
+	output reg user_led11,
+	output reg user_led12,
+	output reg user_led13
```
tcal-x, line 21: We don't need to add these to the pdc file(s)?

MateuszKosmala: Hi, 
Without the user_ledX declaration , I am not able to make a synthesis:
```
Analyzing Verilog file /home/fpga-tool-perf/src/baselitex-nexus/baselitex_nexus.v. VERI-1482
ERROR - synthesis: /home/fpga-tool-perf/src/baselitex-nexus/baselitex_nexus.v(675): user_led0 is not declared. VERI-1128
```

acomodi: It appears that radiant cannot handle auto-inferring of non-declared wires/regs [such cases](https://github.com/chipsalliance/fpga-tool-perf/blob/38bafde58a93eab442e24dbb2f07843704092256/src/baselitex-nexus/baselitex_nexus.v#L674-L693). Without specifying the wires/regs, radiant errors out with the following:

```
`ERROR - synthesis: /home/fpga-tool-perf/src/baselitex-nexus/baselitex_nexus.v(675): user_led0 is not declared. VERI-1128`
```

This was not happening with the yosys-nextpnr-nexus flow as yosys can probably handle this case.

I guess that we do not strictly need to define those as top level ports, but we can just declare them within the module.

acomodi: Given that these leds are not really used, I have removed them from the top IO definition and made them dummy registers

PR #388, opened 2022-01-17
List conda package versions

mkurc-ant:
Added listing actual version of installed conda packages to the Makefile so that they are present in CI logs.


mkurc-ant:
@acomodi Could you review this one ?

acomodi (approved):
LGTM

PR #394, opened 2022-02-04
Properly interpret dates

tgorochowik:
Before:

After:

Precise timestamp can still be checked by hovering over a particular datapoint.


acomodi (approved):
LGTM

PR #396, opened 2022-02-08
Add HPS gen2

acomodi:


acomodi:
@tcal-x Right, the timeout is not yet integrated, I think I can add it to this PR directly, as there seems to be some issues with the exhaust script that does not correctly report how many tests were run and how many have succeded/failed, so I can fix both of them in here, also because those should be not too many changes

PR #400, opened 2022-03-23
Fix CI that generates the result page

mkurc-ant:
Pinned setup-gcloud action to version "v0", fixes the GCS URL link

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@master
+        uses: google-github-actions/setup-gcloud@v0
         with:
           service_account_key: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY_FILE }}
 


.github/workflows/gh-pages.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         run: |
           cd results
           mkdir build-results
-          gsutil cp gs://${{ secrets.GCP_STORAGE_BUCKET }}/*.json.gz build-results
+          gsutil cp ://${{ secrets.GCP_STORAGE_BUCKET }}/*.json.gz build-results
```
acomodi, line 29: Why is this change required?

mkurc-ant: Previously the CI run on push to Antmicro GH complained that this is illegal URL specification: https://github.com/antmicro/fpga-tool-perf/runs/5657033538?check_suite_focus=true

But with this change I have a different error message, see there: https://github.com/antmicro/fpga-tool-perf/runs/5657933114?check_suite_focus=true

So I'm actually puzzled now.

acomodi: Hmm we may try leaving the `gs` prefix there as it was previously working. I think that we can skip the whole CI check and merge it after the change, as this change would not affect the tests at all.

mkurc-ant: Alright, I reverted to the `gs` prefix.

PR #445, opened 2022-07-14
env: use latest packages and start transition to f4pga

acomodi:


tmichalak (approved):
LGTM

PR #456, opened 2022-08-03
ci: limit concurrency

umarcor:
This allows new CI runs to automatically cancel the previous runs corresponding to the same branch. CI is still expected to fail.


mkurc-ant:
LTGM, merging.

PR #495, opened 2022-10-07
Remove redundant log option

robertszczepanski:
This option is included in edalize after its update so it's not necessary here

Signed-off-by: Robert Szczepanski <rszczepanski@antmicro.com>


mkurc-ant:
LGTM.

PR #534, opened 2022-12-19
ci/Suite: revert fb687dcd6b5058c9bf8a2e1e9607a7b257c41ad81

umarcor:


kboronski-ant (changes requested):
LGTM, just change the name of the commit - it does not really tell, what it's for ~~and the hash is not present in main branch's history.~~ EDIT: Sorry - ~~my bad, the hash is there~~, but it's still not exactly a revert.
EDIT 2: Nope, I was right - you accidentally added an extra `1` to the hash. It's fb687dcd6b5058c9bf8a2e1e9607a7b257c41ad8, not fb687dcd6b5058c9bf8a2e1e9607a7b257c41ad81. That commit did deserve to be reverted.

PR #537, opened 2023-01-03
ci: Use verbose script for getting RapidWright JAR link

rw1nkler:
This PR adds a dedicated script for obtaining the RapidWright JAR link. The script has the following advantages over the previous approach:

- retries curl requests on failure
- checks whether the obtained link is valid 
- exits with the error code and additional debug info on failure


mkurc-ant (approved):
LGTM

PR #539, opened 2023-01-03
Remove manual installation of tbb library after updating nextpnr-fpga-interchange conda package

rw1nkler:
After adding tbb library as a run dependency to conda, manual installation of the package in CI is no longer required.

Related to https://github.com/chipsalliance/fpga-tool-perf/pull/536#issuecomment-1369965277


mkurc-ant (approved):
LGTM

PR #551, opened 2023-01-23
ci: pin gcloud to 413.0.0-0

AdamOlech:
This is a temporary workaround for a regression in gsutil which causes multithreaded uploads to fail if the `storage.buckets.get` permission is missing on the caller IAM.
The regression is further documented in the release notes: https://cloud.google.com/storage/docs/release-notes#January_19_2023

This commit should be reverted once the regression has been fixed.

Internal-tag: [#41010]
Signed-off-by: Adam Olech <aolech@antmicro.com>


tmichalak:
LGTM

PR #554, opened 2023-02-01
CI: tie docker image to ubuntu:focal-20221130

kgugala:
The CI is failing to spawn the machines. The issue seems to be the one described in https://github.com/chipsalliance/verible/issues/1663. Locking the image to ubuntu:focal-20221130 should fix the issue


rw1nkler:
LGTM!

PR #555, opened 2023-02-01
Fix nextpnr-fpga_interchange-experimental-single-thread toolchain

rw1nkler:
This PR contains fixes for building the nextpnr-fpga_interchange experimental single-thread toolchain. Without them, the toolchain is built exactly in the same way as the experimental version (with multithreading enabled)


mkurc-ant (approved):
LGTM, merge if green.

PR #556, opened 2023-02-03
Fix usage of lstrip and rstrip

rw1nkler:
Recently we noticed that part of the instability in nextpnr-fpga_interchange toolchains was caused by incorrect usage of `str.lstrip()` and `str.rstrip()` functions during log parsing.

This PR changes the wrong occurrences of those function calls to `removeprefix()` and `removesuffix()` functions respectively. The functions from the python library (`str.removeprefix()`, `str.removesuffix()`) are not used because they were introduced in Python 3.9 and we want to be compatible with older versions of python.

The incorrect usage of `str.lstrip()` and `str.rstrip()` led to incorrect calculations of execution times in the `nextpnr-fpga_interchange` toolchain (most notably in the routing time results).

### DIFF
--- a/toolchains/icestorm.py
+++ b/toolchains/icestorm.py
@@ -24,6 +24,7 @@ import edalize
 
 from toolchains.toolchain import Toolchain
 from utils.utils import Timed, have_exec, get_yosys_resources, get_file_dict
+from utils.utils import removeprefix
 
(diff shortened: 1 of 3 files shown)


tmichalak (approved):
LGTM

PR #557, opened 2023-02-03
Pin type of spawned GCP machines

rw1nkler:
Pin the type of spawned GCP machines to make sure that we use the same type of devices to measure the performance.


tmichalak (approved):
LGTM

Issue #402, opened 2022-04-13
Results visualization: add possibility to hide lines 


acomodi:
At the moment there is no possibility to hide the lines in the charts corresponding to the various toolchains.

![Screenshot from 2022-04-13 09-16-09](https://user-images.githubusercontent.com/44773360/163121328-8df7e344-9973-4bf7-8fe9-9d563f3d6a24.png)

The idea is to be able to click on the colored line belonging to a particular toolchain and disable all the lines corresponding to the particular toolchain.

For context, the [results](https://github.com/chipsalliance/fpga-tool-perf/tree/master/results) directory contains the code to gather and generate the HTML representation of all the results produced by the FPGA tool perf CI.

SiyaoIsHiding:
I would love to contribute to this issue :D. May u assign this issue to me?

SiyaoIsHiding:
@acomodi Just in case u didn't receive the notification

kgugala:
Hi @SiyaoIsHiding I've assigned the issue to you

SiyaoIsHiding:
@kgugala & @acomodi, 
Hello, I am working on this issue.

May I ask what should I put into the environment variable `RESULTS_DIR` when I build the visualization?
The only clue I found is the CI file
```yml
      - name: Use gcloud CLI
        run: |
          cd results
          mkdir build-results
          gsutil cp gs://${{ secrets.GCP_STORAGE_BUCKET }}/*.json.gz build-results
```
But I do not have the access to your gcloud so I still don't know what I should put into this `build-results` folder.

Besides, may you simply give me an example of this `build-results` folder so that I can test the JS codes? I probably only need to modify the JS script inside the result folder, so I do not need to reproduce all your environment.

Thank you for your time and any advice will be appreciated.

Best,
Jane

tmichalak:
Hi @SiyaoIsHiding, thanks for looking into this issue. Let me take a look and get back to you.

tmichalak:
@SiyaoIsHiding so actually for the purpose of solving this issue you could simply generate the data source files using the [exhaust script](https://github.com/chipsalliance/fpga-tool-perf/blob/main/README.md#exhaustive-build), copy the resulting *.json.gz to the created `build-results` directory and run `make preprocess` and `make build` in the [results](https://github.com/chipsalliance/fpga-tool-perf/tree/main/results) directory.
[...]

PR #569, opened 2023-03-22
feat(visual): toggle toolchain visibility in performance visualization

SiyaoIsHiding:
Regarding this issue #402 

First, the `json.gz` file you give me will fail the `make build` process because some records do not have the property `versions` or `family`. Therefore, I added the following codes:
```python
results['versions'] if 'versions' in results else {}, results['family'] if 'family' in results else {}, results['device']
```

Second, I tried to achieve the functionality to toggle the visibility by clicking the LINES instead of the legend, but I failed. I made a workaround to toggle by clicking the legend.

Something like this in the chart configuration won't work
```js
interaction: {
        mode: 'index',
        intersect: false,
        axis: 'x',
        onHover: function (event, chartElement) {
          console.log("hover"); // no log
        },
        onClick: function (event, chartElement) {
          console.log("click");  // no log either
        }
      }
```

Something like this won't work either
```js
function clickHandler(click){
    const points = myChart.getElementsAtEventForMode(click, 'nearest', { intersect: true}, true);
    console.log(points); // always an empty array
}
myChart.canvas.onclick = clickHandler;
```

Hope you find the workaround acceptable.

### DIFF
--- a/results/html/index.html
+++ b/results/html/index.html
@@ -76,6 +76,7 @@
 <script charset="utf-8">
 
 var visible_elem = null;
+var visible_chart_ids = [];
 
 // sets visibility of an element with the w3-hide class by appending/removing w3-show class
 function setVisibility(elem, on) {
@@ -183,7 +184,8 @@ function selectStat(elem_id, device, project) {
 }
 
 function LineChart(canvas_name, title, unitstr, labels, datasets, legend) {
-  return new Chart(document.getElementById(canvas_name), {
+    visible_chart_ids.push(canvas_name);
+  const myChart = new Chart(document.getElementById(canvas_name), {
     type: 'line',
     data: {
       labels: labels,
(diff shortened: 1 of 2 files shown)


tmichalak:
@SiyaoIsHiding thanks, I will take a look. A quick comment would be to fix the formatting errors reported by the [CI](https://github.com/chipsalliance/fpga-tool-perf/actions/runs/4486438440/jobs/7895984833?pr=569) .

SiyaoIsHiding:
> @SiyaoIsHiding thanks, I will take a look. A quick comment would be to fix the formatting errors reported by the [CI](https://github.com/chipsalliance/fpga-tool-perf/actions/runs/4486438440/jobs/7895984833?pr=569) .

Thx for your advice. It should be able to pass the yapf format check now.

SiyaoIsHiding:
@tmichalak 
He says "disable ALL the lines corresponding to the particular toolchain." I assume he means disable the corresponding line in all the graphs. 
If you want to only select/deselect a single line, that would be easy to achieve. 
May I confirm what you want?

Besides, I would not be able to achieve the functionality that triggered by clicking a data point instead of a legend. I tried and I failed, as explained above.

